PR #36, opened 2022-08-16
test: add a scenario with revoke after the lock period ends in buOLAS

kupermind:
- Adding a test where `revoke()` is called after the `buOLAS` lock period ends, and `createLockFor()` becomes available in this scenario.

### DIFF
--- a/test/buOLAS.js
+++ b/test/buOLAS.js
@@ -362,17 +362,22 @@ describe("buOLAS", function () {
             await bu.connect(owner).createLockFor(account.address, oneOLASBalance, numSteps);
 
             // Move three years in time
+            // oneOLASBalance is fully unlocked at this point
             ethers.provider.send("evm_increaseTime", [3 * oneYear + 100]);
             ethers.provider.send("evm_mine");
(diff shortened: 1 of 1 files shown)


77ph (approved):

test/buOLAS.js
```diff
@@ -347,6 +349,39 @@ describe("buOLAS", function () {
...
+            ethers.provider.send("evm_mine");
+            // Revoke at this point of time (which must never be performed in real situations)
+            await bu.connect(owner).revoke([account.address]);
+
+            // Now createLockFor for the same account
+            await bu.connect(owner).createLockFor(account.address, 3, numSteps);
+
+            // Try to withdraw right away with the calculated releasable amount
+            let amount = await bu.releasableAmount(account.address);
+            // Due to the underflow with "unchecked" statement in the uint256 value, the amount will be that much:
```
mariapiamo, line 375:
```suggestion
            // The amount would be -10^18.  However, due to the "unchecked" statement in the uint256 
            // the amount would become
```

DavidMinarsch (approved):
LGTM

PR #43, opened 2023-01-05
chore: add test contract deployment

kupermind:
- Initial deployment of contracts with the deployer being the owner and the minter.


DavidMinarsch:
Stale

PR #48, opened 2023-02-22
feat and test: implementing wrapped veOLAS

kupermind:
- Implementing wrapped veOLAS;
- Adding tests;
- Coverage is 100%.

### DIFF
--- a/contracts/wveOLAS.sol
+++ b/contracts/wveOLAS.sol
@@ -95,11 +95,10 @@ contract wveOLAS {
     /// @param idx User point number.
     /// @return uPoint The requested user point.
     function getUserPoint(address account, uint256 idx) public view returns (PointVoting memory uPoint) {
-        if (account != address(0)) {
-            uint256 userNumPoints = IVEOLAS(ve).getNumUserPoints(account);
-            if (userNumPoints > 0) {
-                uPoint = IVEOLAS(ve).getUserPoint(account, idx);
-            }
+        // Get the number of user points
+        uint256 userNumPoints = IVEOLAS(ve).getNumUserPoints(account);
+        if (userNumPoints > 0) {
+            uPoint = IVEOLAS(ve).getUserPoint(account, idx);
         }
     }
 
--- a/contracts/wveOLAS.sol
+++ b/contracts/wveOLAS.sol
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+// Structure for veOLAS points
+struct PointVoting {
+    int128 bias;
+    int128 slope;
+    uint64 ts;
+    uint64 blockNumber;
+    uint128 balance;
+}
+
+interface IVEOLAS {
+    /// @dev Gets the total number of supply points.
+    /// @return numPoints Number of supply points.
+    function totalNumPoints() external view returns (uint256 numPoints);
+
+    /// @dev Gets the supply point of a specified index.
+    /// @param idx Supply point number.
+    /// @return sPoint Supply point.
+    function mapSupplyPoints(uint256 idx) external view returns (PointVoting memory sPoint);
+
+    /// @dev Gets the number of user points.
+    /// @param account Account address.
+    /// @return userNumPoints Number of user points.
+    function getNumUserPoints(address account) external view returns (uint256 userNumPoints);
+
+    /// @dev Gets the checkpoint structure at number `idx` for `account`.
+    /// @notice The out of bound condition is treated by the default code generation check.
+    /// @param account User wallet address.
+    /// @param idx User point number.
+    /// @return uPoint The requested user point.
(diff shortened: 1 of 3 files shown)


test/wveOLAS.js
```diff
@@ -0,0 +1,509 @@
+/*global describe, context, beforeEach, it*/
+
+const { expect } = require("chai");
+const { ethers } = require("hardhat");
+
+describe("Wrapped Voting Escrow OLAS", function () {
```
kupermind, line 6: Tests are reduced in the next PR, since there is no need to repeat ve-only tests.

contracts/wveOLAS.sol
```diff
@@ -0,0 +1,170 @@
...
+error WrongTimestamp(uint256 minTimeStamp, uint256 providedTimeStamp);
+
+/// @dev Called function is implemented in a specified veOLAS contract.
+/// @param ve Original veOLAS address.
+error ImplementedIn(address ve);
+
+/// @title wveOLAS - Wrapper smart contract for view functions of veOLAS contract
+/// @author AL
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+contract wveOLAS {
```
kupermind, line 78: Functions are wrapped exactly as described in the specification doc.

test/wGovernanceOLAS.js
```diff
@@ -0,0 +1,424 @@
...
+            const proposers = [proxyAddress];
+            const Timelock = await ethers.getContractFactory("Timelock");
+            const timelock = await Timelock.deploy(minDelay, proposers, executors);
+            await timelock.deployed();
+            // console.log("Timelock deployed to", timelock.address);
+
+            // Deploy Governance Bravo
+            const GovernorBravo = await ethers.getContractFactory("GovernorOLAS");
+            const governor = await GovernorBravo.deploy(wve.address, timelock.address, initialVotingDelay,
+                initialVotingPeriod, initialProposalThreshold, quorum);
```
kupermind, line 93: Governance based on the veOLAS wrapper address.

contracts/wveOLAS.sol
kupermind: All the functions needed by the Governor contract are implemented along with others affected.

77ph (approved):

test/wGovernanceOLAS.js
```diff
@@ -0,0 +1,424 @@
+/*global describe, context, beforeEach, it*/
+
+const { expect } = require("chai");
```
kupermind, line 3: New test files are basically clones of original ones, with the veOLAS wrapper being called instead of veOLAS where applicable

contracts/wveOLAS.sol
```diff
@@ -0,0 +1,170 @@
...
+    /// @notice The requested block number must be at least equal to the zero supply point block number.
+    /// @param blockNumber Block number to calculate the total voting power at.
+    /// @return vPower Total voting power.
+    function getPastTotalSupply(uint256 blockNumber) external view returns (uint256 vPower) {
+        // Get the zero supply point
+        PointVoting memory sPoint = IVEOLAS(ve).mapSupplyPoints(0);
+        // Check the requested block number to be at least equal to the zero supply point block number
+        if (blockNumber >= sPoint.blockNumber) {
+            vPower = IVEOLAS(ve).getPastTotalSupply(blockNumber);
+        }
```
mariapiamo, line 163: If the `blockNumber` is not smaller than the very first supply point (e.g. the zero supply point), the `getPastTotalSupply` veOLAS original behavior is correct. So the latter is called.
Otherwise, we return zero instead reverting without mentioning the reasons ( changing the native behavior of the original Curve Voting Escrow that plain reverts in this case)

mariapiamo (commented):
1. Please go over my comments. Once you agree with them, I'll use them for the doc file I am going to create
2. Could you please recall me if we have a test on veOLAS that shows the issues on the view function we are solving here?
3. Still `npm run fork` command does not work here. But this is solved in the next branch

contracts/wveOLAS.sol
```diff
@@ -0,0 +1,170 @@
...
+
+interface IVEOLAS {
+    /// @dev Gets the total number of supply points.
+    /// @return numPoints Number of supply points.
+    function totalNumPoints() external view returns (uint256 numPoints);
+
+    /// @dev Gets the supply point of a specified index.
+    /// @param idx Supply point number.
+    /// @return sPoint Supply point.
+    function mapSupplyPoints(uint256 idx) external view returns (PointVoting memory sPoint);
```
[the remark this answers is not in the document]
kupermind: These are state variable analogues from the original veOLAS contract.

kupermind:
@mariapiamo you are correct on 2 - I will explicitly create two tests in veOLAS and wveOLAS test files to show that the view function returns incorrect value in the first one and the correct one in the second one.

PR #51, opened 2023-02-26
chore and test: updating deployment scripts

kupermind:
- updating deployment scripts with wveOLAS and deprecated Sale;
- updating README-s;
- updating tests and deployment script where applicable.


DavidMinarsch (commented):
The graphic should also contain wveOLAS

test/wveOLAS.js
```diff
@@ -287,14 +287,19 @@ describe("Wrapped Voting Escrow OLAS", function () {
...
             await ve.connect(owner).increaseAmount(oneOLABalance);
             await ve.connect(owner).increaseAmount(oneOLABalance);
-            let blockNumber = await ethers.provider.getBlockNumber("latest");
 
-            // Get past votes of the owner
-            const votesOwner = ethers.BigNumber.from(await wve.getPastVotes(owner.address, blockNumber));
-            expect(Number(votesOwner)).to.greaterThan(0);
+            // The past votes before the lock must be zero
+            let votesOwner = ethers.BigNumber.from(await wve.getPastVotes(owner.address, blockNumber));
+            expect(votesOwner).to.equal(0);
```
kupermind, line 297: Show that the wrapped function is correctly returning zero voting power when asked for the block number before the first lock.

test/veOLAS.js
```diff
@@ -439,9 +439,9 @@ describe("Voting Escrow OLAS", function () {
             await ve.connect(owner).increaseAmount(oneOLABalance);
             await ve.connect(owner).increaseAmount(oneOLABalance);
 
-            // Get past votes of the owner
+            // Get past votes of the owner (bug resolved in wveOLAS)
             const votesOwner = await ve.getPastVotes(owner.address, blockNumber);
-            expect(Number(votesOwner)).to.greaterThan(0);
+            expect(votesOwner).to.greaterThan(0);
```
kupermind, line 444: Show that the original function is incorrectly returning non-zero voting power when asked for the block number before the first lock.

77ph (approved):

PR #78, opened 2023-06-23
Create script for CM removal

mariapiamo:
Script providing the inputs for CM removal proposal

### DIFF
--- a/scripts/deployment/deploy_19_revoke_CM.js
+++ b/scripts/deployment/deploy_19_revoke_CM.js
@@ -8,7 +8,6 @@ async function main() {
     const dataFromJSON = fs.readFileSync(globalsFile, "utf8");
     let parsedData = JSON.parse(dataFromJSON);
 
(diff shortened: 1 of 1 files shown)


scripts/deployment/deploy_19_revoke_CM.js
```diff
@@ -0,0 +1,43 @@
...
+
+const { ethers } = require("hardhat");
+
+async function main() {
+    const fs = require("fs");
+    const globalsFile = "globals.json";
+    const dataFromJSON = fs.readFileSync(globalsFile, "utf8");
+    let parsedData = JSON.parse(dataFromJSON);
+
+    const adminRole = ethers.utils.id("TIMELOCK_ADMIN_ROLE");
```
kupermind, line 11: This line is not needed since the admin role was already revoked, and is not used later in the script

mariapiamo: right! Thanks

mariapiamo, line 11:
```suggestion
```

PR #79, opened 2023-08-17
chore: scripts to audit on-chain contract setup

kupermind:
- Adding scripts to audit on-chain contract setup.


abis/0.8.19/GovernorOLAS.json
```diff
@@ -1,1425 +0,0 @@
-{
```
kupermind, line 1: This ABI is not used anywhere in contracts or setup, it was the intermediary one

hardhat.config.js
```diff
@@ -106,7 +106,7 @@ module.exports = {
                 settings: {
                     optimizer: {
                         enabled: true,
-                        runs: 750,
+                        runs: 1000000,
```
kupermind, line 109: Restoring the justice!

scripts/audit_chains/audit_contracts_setup.js
```diff
@@ -0,0 +1,385 @@
...
+    const timelock = await governor.timelock();
+    customExpect(timelock, globalsInstance["timelockAddress"], log + ", function: timelock()");
+
+    // Check version, hardcoded
+    const version = await governor.version();
+    customExpect(version, "1", log + ", function: version()");
+
+    // Check quorumNumerator, hardcoded
+    const quorumNumerator = await governor["quorumNumerator()"]();
+    customExpect(quorumNumerator.toString(), globalsInstance["quorum"], log + ", function: quorumNumerator()");
```
mariapiamo, line 211: I would have expected that this and the the following two would have shown an issue. Indeed governorOLAS on goerli have different `quorum`, `initialVotingDelay`,  `initialVotingPeriod`. I guess that hardcoded values should be used here

kupermind: We compare values with our globals files - they are different for mainnet and goerli. If we want this to be exactly as on the mainnet, we need to use its values then.

kupermind: Updated with the mainnet values.

mariapiamo (approved):

PR #83, opened 2023-08-28
feat: initial CM guard implementation

kupermind:
- CM guard implementation;
- Added tests.

### DIFF
--- a/contracts/multisigs/GuardCM.sol
+++ b/contracts/multisigs/GuardCM.sol
@@ -7,6 +7,18 @@ interface IGovernor {
     function state(uint256 proposalId) external returns (ProposalState);
 }
 
+// Governor proposal state
+enum ProposalState {
+    Pending,
+    Active,
+    Canceled,
+    Defeated,
+    Succeeded,
+    Queued,
+    Expired,
+    Executed
+}
+
 /// @dev Only `owner` has a privilege, but the `sender` was provided.
 /// @param sender Sender address.
 /// @param owner Required sender address as an owner.
@@ -44,17 +56,10 @@ error NoSelfCall();
 /// @param selector Function selector.
 error NotAuthorized(address target, bytes4 selector);
 
-// Governor proposal state
-enum ProposalState {
-    Pending,
-    Active,
-    Canceled,
-    Defeated,
-    Succeeded,
-    Queued,
-    Expired,
-    Executed
-}
+/// @dev The proposal is not yet defeated.
+/// @param proposalId Proposal Id.
+/// @param state Current proposal state.
+error NotYetDefeated(uint256 proposalId, ProposalState state);
 
 /// @title GuardCM - Smart contract for Gnosis Safe community multisig (CM) guard
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
@@ -268,6 +273,8 @@ contract GuardCM {
             ProposalState state = IGovernor(governor).state(governorCheckProposalId);
             if (state == ProposalState.Defeated) {
                 paused = 2;
+            } else {
+                revert NotYetDefeated(governorCheckProposalId, state);
             }
         } else {
             // msg.sender is not a timelock, nor a multisig
--- a/contracts/multisigs/GuardCM.sol
+++ b/contracts/multisigs/GuardCM.sol
@@ -56,10 +56,10 @@ error NoSelfCall();
 /// @param selector Function selector.
 error NotAuthorized(address target, bytes4 selector);
 
-/// @dev The proposal is not yet defeated.
+/// @dev The proposal is not defeated.
 /// @param proposalId Proposal Id.
 /// @param state Current proposal state.
-error NotYetDefeated(uint256 proposalId, ProposalState state);
+error NotDefeated(uint256 proposalId, ProposalState state);
 
 /// @title GuardCM - Smart contract for Gnosis Safe community multisig (CM) guard
 /// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
@@ -146,18 +146,16 @@ contract GuardCM {
 
         // Prepare the decoding data sets
         address[] memory targets;
-        uint256[] memory values;
         bytes[] memory callDatas;
         if (selector == SCHEDULE) {
             targets = new address[](1);
-            values = new uint256[](1);
             callDatas = new bytes[](1);
             // Decode the data in the schedule function
-            (targets[0], values[0], callDatas[0], , , ) =
+            (targets[0], , callDatas[0], , , ) =
                 abi.decode(payload, (address, uint256, bytes, bytes32, bytes32, uint256));
         } else {
             // Decode the data in the scheduleBatch function
-            (targets, values, callDatas, , , ) =
+            (targets, , callDatas, , , ) =
             abi.decode(payload, (address[], uint256[], bytes[], bytes32, bytes32, uint256));
         }
 
@@ -267,7 +265,7 @@ contract GuardCM {
             if (state == ProposalState.Defeated) {
                 paused = 2;
             } else {
-                revert NotYetDefeated(governorCheckProposalId, state);
+                revert NotDefeated(governorCheckProposalId, state);
             }
         } else {
             // msg.sender is not a timelock, nor a multisig
--- a/contracts/multisigs/GuardCM.sol
+++ b/contracts/multisigs/GuardCM.sol
@@ -3,6 +3,10 @@ pragma solidity ^0.8.21;
 
 import {Enum} from "@gnosis.pm/safe-contracts/contracts/common/Enum.sol";
 
+interface IGovernor {
+    function state(uint256 proposalId) external returns (ProposalState);
+}
+
 /// @dev Only `owner` has a privilege, but the `sender` was provided.
 /// @param sender Sender address.
 /// @param owner Required sender address as an owner.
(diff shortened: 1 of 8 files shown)


contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,301 @@
...
+    mapping(uint256 => bool) public mapAllowedTargetSelectors;
+
+    // Schedule selector
+    bytes4 public constant SCHEDULE = bytes4(keccak256(bytes("schedule(address,uint256,bytes,bytes32,bytes32,uint256)")));
+    // ScheduleBatch selector
+    bytes4 public constant SCHEDULE_BATCH = bytes4(keccak256(bytes("scheduleBatch(address[],uint256[],bytes[],bytes32,bytes32,uint256)")));
+    // Initial check governance proposal Id
+    // Calculated from the proposalHash function of the GovernorOLAS with the following parameters:
+    // targets = [address(0)], values = [0], calldatas = [0x], description = "Is governance alive?"
+    uint256 public governorCheckProposalId = 62151151991217526951504761219057817227643973118811130641152828658327965685127;
```
kupermind, line 93: This is a modifiable proposal Id that can be checked on to understand if the governance is alive

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+contract GuardCM {
+    event GovernorUpdated(address indexed governor);
+    event SetTargetSelectors(address[] indexed targets, bytes4[] indexed selectors, bool[] statuses);
+    event GovernorCheckProposalIdChanged(uint256 indexed proposalId);
+    event GuardPaused(address indexed account);
+    event GuardUnpaused();
+
+    // Owner address
+    address public immutable owner;
```
mariapiamo, line 75: what if we need to update the Timelock? Can we make this mutable via owner?

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+    // Mapping of address + bytes4 selector => enabled / disabled
+    mapping(uint256 => bool) public mapAllowedTargetSelectors;
+
+    // Schedule selector
+    bytes4 public constant SCHEDULE = bytes4(keccak256(bytes("schedule(address,uint256,bytes,bytes32,bytes32,uint256)")));
+    // ScheduleBatch selector
+    bytes4 public constant SCHEDULE_BATCH = bytes4(keccak256(bytes("scheduleBatch(address[],uint256[],bytes[],bytes32,bytes32,uint256)")));
+    // Initial check governance proposal Id
+    // Calculated from the proposalHash function of the GovernorOLAS with the following parameters:
+    // targets = [address(0)], values = [0], calldatas = [0x], description = "Is governance alive?"
```
mariapiamo, line 92: We cannot have more than one governance proposal with all the same data.
So we should have
// targets = [address(0)], values = [0], calldatas = [0x], description = "Is governance alive - test number x?" ,
where x increases with the increase of this kind of proposal

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+    uint256 public governorCheckProposalId = 62151151991217526951504761219057817227643973118811130641152828658327965685127;
+
+    /// @dev GuardCM constructor.
+    /// @param _timelock Timelock address.
+    /// @param _multisig Community multisig address.
+    /// @param _governor Governor address.
+    constructor(address _timelock, address _multisig, address _governor) {
```
mariapiamo, line 99: This is going to be updated by using the initial approved methods as construct inputs, correct?

contracts/multisigs/GuardCM.sol
mariapiamo: If yes, change owner method has to be added

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+            revert ZeroAddress();
+        }
+
+        governor = newGovernor;
+        emit GovernorUpdated(newGovernor);
+    }
+
+    /// @dev Changes the governor check proposal Id.
+    /// @param proposalId Governor check proposal Id.
+    function changeGovernorCheckProposalId(uint256 proposalId) external {
```
mariapiamo, line 123: Is this the correct workflow: 
- CM make a proposal on governor (so proposer role is also required) using the parameters suggested above such that the above proposalID is returned
- If the proposal does not succeed the CM have the ability to pause the guard
- Once the guard is paused, CM has fully executor role, so it can schedule and execute via timelock the change of the proposalId

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+            revert OwnerOnly(msg.sender, owner);
+        }
+
+        // Check for the zero value
+        if (proposalId == 0) {
+            revert ZeroValue();
+        }
+
+        governorCheckProposalId = proposalId;
+        emit GovernorCheckProposalIdChanged(proposalId);
```
mariapiamo, line 134: can we also somehow emit also the proposal description?

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+
+            // Check the authorized combination of target and selector
+            if (!mapAllowedTargetSelectors[targetSelector]) {
+                revert NotAuthorized(targets[i], bytes4(callDatas[i]));
+            }
+        }
+    }
+
+    /// @dev Checks the transaction for authorized arguments.
+    /// @notice Scheduling in timelock in checked against authorized targets and signatures.
```
mariapiamo, line 178:
```suggestion
    /// @notice Scheduling in timelock is checked against authorized targets and signatures.
```

contracts/multisigs/GuardCM.sol
```diff
@@ -0,0 +1,292 @@
...
+        address payable,
+        bytes memory,
+        address
+    ) external view {
+        // Just return if paused
+        if (paused == 1) {
+            // Call to the timelock
+            if (to == owner) {
+                // No delegatecall is allowed
+                if (operation == Enum.Operation.DelegateCall) {
```
mariapiamo, line 201: why not delegate call or seld-multisig call?

contracts/multisigs/GuardCM.sol
77ph: "Is governance alive?"
vs
"Is governance alive??"
vs
"Is governance alive - test number x?"
since it's a hash of a human-readable string, the text of the pattern itself doesn't matter.
You just need to agree on what text will be in next "check alive" proposalId. And change it. See above.
descriptionHash (bytes32 which itself is the keccak256 hash of the description string)
uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);

contracts/multisigs/GuardCM.sol
77ph: This is a one-way hash.
If the description is understood as text, then it is also included as a hash.

contracts/multisigs/GuardCM.sol
mariapiamo: yeap. Agree, I was proposing a standard format we can use for this :-)

contracts/multisigs/GuardCM.sol
kupermind: We do not believe there is a scenario where we need to update the Timelock. If that's the case, we need to re-deploy and re-set the guard. This would involve changing the module as well.

contracts/multisigs/GuardCM.sol
77ph: - delegatecall to timelock does not make practical sense, but we do not need storage collisions when timelock can be used as a library to change storage of Guard.
- self-multisig not allowed for change owners, thershould, etc

contracts/multisigs/GuardCM.sol
kupermind: Yeah, we can modify that at any time once the first proposal is used.

contracts/multisigs/GuardCM.sol
kupermind: Correct, this will update regardless of what was set before.

77ph (approved):

contracts/multisigs/GuardCM.sol
kupermind: Because we don't want the multisig to change itself. And the delegatecall... is super unexpected here when calling the Timelock. This means that the multisig wants to joint something in the storage with the Timelock - super weird ).

contracts/multisigs/GuardCM.sol
mariapiamo: In that case we need also to have a governance vote that pause guard and remove the timelock module form the CM then

contracts/multisigs/GuardCM.sol
mariapiamo: why not simply keeping the CM proposer role?

kupermind: The proposal is done via the governor, as we are checking the governor itself. In order to propose in the governor contract, one needs veOLAS

mariapiamo: yeah right, the proposer role of the CM is for scheduling on the timelock

77ph (approved):

PR #85, opened 2023-08-30
chore: deployment script for community multisig

kupermind:
- Deployment scripts for the community multisig.


77ph (approved):

scripts/proposals/proposal_03_revoke_CM.js
```diff
@@ -16,7 +16,7 @@ async function main() {
     const CMAddress = parsedData.CM;
 
     // Obtaining proposal values
-    console.log("19. Revoking CM in the Timelock");
+    console.log("3. Revoking CM in the Timelock");
```
mariapiamo, line 19: none of this will be done since we are using the guard. Maybe it is less confusing if we remove this, no?

kupermind: Let it be in the proposals, it is a good practice to have something for all the contracts.

PR #156, opened 2024-10-14
chore: preparing for Mode deployment

kupermind:
- Preparing for Mode deployment.


77ph (approved):

77ph (approved):

DavidMinarsch (approved):
Any reason we updated the abis? Would it not be desirable to have identical ones across networks?

kupermind:
@DavidMinarsch we just follow the process there. The only difference in those ABIs is the bytecode, as the contract is compiled with a more recent compiler version. This lets us have contracts untouched while continue to possibly optimize due to compiler advancements.

PR #159, opened 2024-10-28
feat: OLAS burner contract

kupermind:
- Adding OLAS burner contract.


DavidMinarsch (approved):

contracts/Burner.sol
```diff
@@ -0,0 +1,66 @@
...
+    constructor(address _olas) {
+        if (_olas == address(0)) {
+            revert ZeroAddress();
+        }
+
+        olas = _olas;
+    }
+
+    /// @dev Burns OLAS tokens owned by this contract.
+    function burn() external {
```
77ph, line 44: https://github.com/valory-xyz/autonolas-governance/blob/main/contracts/OLAS.sol#L118
What is the meaning of this contract, anyone can call this function for themselves.

77ph: OK

PR #163, opened 2024-11-06
feat: wormhole relayer for arbitrary msg.value-s

kupermind:
- Wormhole relayer for arbitrary `msg.value`-s.


kupermind:
This is the solution for any governance proposed message passing via Wormhole.

77ph (dismissed):
I agree with this idea. It makes sense to add reentrancy protection as usual.

77ph (approved):
LGTM

DavidMinarsch (approved):

PR #210, opened 2026-06-29
audits: internal19 — full governance security re-audit (PASS-WITH-FINDINGS)

77ph:
## Internal security re-audit — `audits/internal19`

A comprehensive, from-scratch re-audit of all **27 production contracts** on `main` (`bae8da6`), independent of prior internal reviews.

**Verdict: PASS-WITH-FINDINGS — 0 Critical / 0 High / 0 Medium / 7 Low / 16 Info.**

No defect was found by which an unprivileged actor can subvert the live governance system, mint or move funds, or cause an unauthorized cross-chain execution.

### Headline findings (Low)
- **veOLAS** raw `getPastVotes` omits the pre-first-checkpoint guard that `balanceOfAt`/`totalSupplyAt` and the `wveOLAS` wrapper all carry → the token's own `IVotes` surface is non-self-safe for a direct integrator (the deployed Governor is protected because it reads through `wveOLAS`).
- **VoteWeighting** `removeNominee` leaves a dangling slope → weighting skew, uncapped `relativeWeight > 1e18`, and an owner-scoped liveness revert; the last-element guard also leaves a dangling `mapNomineeIds`.
- **buOLAS** owner-gated `revoke`/`createLockFor` accounting is fragile to mis-sequencing (overwrite of `transferredAmount`, no idempotency guard, no slot reset on re-lock, deferred burn) → an owner mistake can strand burns or brick a beneficiary's lock.
- The cross-chain verifiers authorize `(target, selector, chainId)` but **do not bound the inner native `value`** the L2 executor forwards.
- `veOLAS._checkpoint` 255-week history-fill truncation (corrupts global vote accounting only after a ~4.9-year dormancy; withdrawals stay safe).

### Two cross-chain `>Medium` candidates — resolved decisively
- **Verifier-vs-executor parser divergence: not a finding.** The verifier and every L2 executor segment each bridged buffer into byte-identical records (cursor `20 → +16 → +payloadLength` vs `20 → +12 → +4 → +payloadLength`; `20+16 == 20+12+4 == 36`), and the selector is checked on the exact payload that executes. No unauthorized call is reachable.
- **Inner `value` not bound by the allowlist: real but Low.** Trigger is the threshold-trusted Community Multisig, destination is an allowlisted target, amount is balance-bounded.

### Architecture note
The CM holds `PROPOSER`+`EXECUTOR`(+`CANCELLER`) directly on the Timelock and can bypass `governorDelay`; the sole containment is the `GuardCM` allowlist. This is intentional, but the practical security boundary of the whole system reduces to CM-signer honesty plus correct `GuardCM` allowlist curation — a conscious, monitored invariant (keep Timelock `minDelay > 0`).

Full report, clean-coverage note, and fixes: `audits/internal19/README.md`.

— audit-claude

### DIFF
--- a/audits/internal19/README.md
+++ b/audits/internal19/README.md
@@ -1,22 +1,22 @@
 # Autonolas Governance — Internal Security Re-Audit (internal19)
 
 The review was performed on the contract code in this repository on the `main` branch
-(commit `bae8da6`; the full 27-contract project as of the current `main`).
+(commit `bae8da6`).
 
-This is a comprehensive, from-scratch re-audit of the complete `autonolas-governance` contract set,
+This is a comprehensive, from-scratch re-audit of the in-scope `autonolas-governance` contract set,
 independent of prior internal reviews. The goal was a thorough adversarial assessment of every
-production contract, with particular attention to the historical-view (vote-escrow) surfaces, the
-gauge-weighting accounting, the Community-Multisig guard stack, and the cross-chain governance
+in-scope production contract, with particular attention to the historical-view (vote-escrow) surfaces,
+the gauge-weighting accounting, the Community-Multisig guard stack, and the cross-chain governance
 execution path.
 
-**Overall verdict: PASS-WITH-FINDINGS — 0 Critical / 0 High / 0 Medium / 7 Low / 16 Info.**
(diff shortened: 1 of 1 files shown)


77ph:
Closing this internal review. We are replacing it with a from-scratch, manual, exhaustive re-audit of the governance contract set; a fresh report will follow under audits/.

PR #124, opened 2024-03-06
refactor and test: addressing internal audit and adding full coverage

kupermind:
- Addressing internal audit with small refactoring;
- Adding full test coverage.

PR #212, opened 2026-07-03
scripts + tests: veto-stack deploy (Task 1) + mainnet-fork verification

kupermind:
## Summary

Task 1 of the two-layer governance-delay defence: adds a cancel-only Veto-Governor bound to a dedicated Veto Timelock. **No new contracts** — both are redeploys of the existing audited `contracts/Timelock.sol` and `contracts/GovernorOLAS.sol` with different constructor parameters. Same audit surface.

### Deploy scripts

Three forge-based shell scripts, style matches `deploy_27_governor.sh`. Run in strict order:

- **`deploy_28_veto_timelock.sh`** — deploys Veto Timelock empty (no proposers/executors; the Veto-Governor doesn't exist yet). Deployer holds `TIMELOCK_ADMIN_ROLE`, released only in `deploy_30`.
- **`deploy_29_veto_governor.sh`** — deploys Veto-Governor bound to Veto Timelock with `vetoGovernorDelay = 0` (`globals.vetoGovernorDelay`). On-chain sanity checks:
  - `veto.token() == mainGovernor.token()` — must be the **wveOLAS wrapper `0x4039…`**, not raw veOLAS `0x7e01…`. Aborts if mismatched.
  - `globals.vetoVotingDelay` must equal the live main `Governor.votingDelay()`. Copying a raised Layer-1 value would push the veto cycle beyond the 14 d `governorDelay` window; aborts if divergent.
- **`deploy_30_wire_freeze_veto_timelock.sh`** — wires + role-freezes Veto Timelock in this exact order:
  1. `grantRole(PROPOSER_ROLE, vetoGov)`
  2. `grantRole(EXECUTOR_ROLE, vetoGov)`
  3. `revokeRole(TIMELOCK_ADMIN_ROLE, vetoTimelockAddress)` — closes the self-administration hole (an unfrozen VT would let ONE won veto vote grant an attacker standing `PROPOSER` — permanent, voteless griefing of every Timelock A op).
  4. `renounceRole(TIMELOCK_ADMIN_ROLE, deployer)`

  **CANCELLER on VT is intentionally NOT granted.** Its only consumer would be `GovernorCompatibilityBravo.cancel(uint256)` reaching `_timelock.cancel` — unreachable at `vetoGovernorDelay = 0`. The CANCELLER role the veto stack DOES need is on **Timelock A** (Task 2 grant), so VT can call `A.cancel(badId)`.

  Post-freeze verification asserts the exact final role state; script aborts red on any deviation.

### `globals_mainnet.json` additions

```
"vetoMinDelay":          "0",         // Veto Timelock's minDelay
"vetoGovernorDelay":     "0",         // Veto-Governor's governorDelay
"vetoVotingDelay":       "13091",     // today's ~1.82 d — NOT raised Layer-1
"vetoVotingPeriod":      "19636",
"vetoProposalThreshold": "5000000000000000000000",
"vetoQuorum":            "3"
```

Plus placeholder keys `vetoTimelockAddress` + `vetoGovernorAddress` populated by the scripts.

### Fork-test suite (`test/forge/ForkGovernanceVetoDelay.t.sol`)

12-test mainnet-fork verification of the mechanism (setUp matches the iterative-wiring variant of the deploy scripts, including the role-freeze and the intentional CANCELLER-omission):

- **T1** — `governorDelay` bump blocks execute until D elapses
- **T2** — same-Timelock cancel loses to bad op (motivation for a separate VT)
- **T3** — veto through VT cancels a queued A op with 0 delay
- **T5** — VT is powerless beyond cancel (VT → A grantRole reverts)
- **T5b** — post-freeze VT cannot re-role itself (role-freeze holds)
- **T6** — self-anchoring: `updateGovernorDelay(0)` is delayed AND vetoable
- **T8** — Layer 1 `votingDelay` lever engages
- **T8b** — brake removability under active griefing (unopposed revocation succeeds via main Governor)
- **T9** — veto-stack self-disarm bounded (`VT.updateDelay(1)` bricks Layer 2 only; main stack + Layer 1 unaffected)
- **ATTACK** — griefing mechanism proof (veto can cancel any pending A op)
- **HOLE** — pre-freeze escalation proof (an unfrozen VT lets one veto vote mint a standing PROPOSER — proves the role-freeze is required)
- **precondition** — `A.cancel` requires `isOperationPending`; retry is safe

## End-to-end simulation (anvil mainnet fork, block ~25,453k)

- **`deploy_28`** — VT deployed at `0x59de…AB01`; `deployer + VT` both hold `TIMELOCK_ADMIN_ROLE`; `minDelay = 0`. ✓
- **`deploy_29`** — Veto-Governor at `0xE85F…C1B7`; `token/timelock/votingDelay/governorDelay` all as expected; token-match assertion PASS. ✓
- **`deploy_30`** — 4 txs land; post-freeze verification: `PROPOSER=true, EXECUTOR=true, CANCELLER=false, ADMIN(VT)=false, ADMIN(deployer)=false, minDelay=0`. ✓
- **End-to-end** — after pranking Task 2's `A.grantRole(CANCELLER, VT)` and scheduling a bad op via Governor A, pranked VT to `A.cancel(badId)`. Bad op: `isOperationPending → false`, `isOperation → false`. ✓

## Test plan
- [x] `bash -n` syntax check on all three scripts.
- [x] `jq empty globals_mainnet.json` — JSON valid.
- [x] `forge test --mc ForkGovernanceVetoDelay` — **12/12 PASS** on live mainnet fork.
- [x] Pre-existing `test/forge/ForkDeployGovernance.t.sol` — **31/31 still PASS**.
- [x] ESLint on `scripts/deployment/` — no new errors (3 pre-existing warnings in unrelated bridge files).
- [x] Solhint on `contracts/*` — clean.
- [ ] Dry-run on Sepolia before mainnet — recommended before Task 2.


DavidMinarsch:
## Review — deploy scripts vs the merged rollout spec (autonolas-analytics PR #120)

Ran the full verification myself: checked every script line against the merged rollout doc (`governance/olas-governance-delay-rollout-2026-07-03.md`), against `contracts/Timelock.sol` / `contracts/GovernorOLAS.sol` constructors, against live mainnet state, and **ran the PR's fork suite against a live mainnet fork: 12/12 PASS**.

### Verified correct (independently)

- **Deploy order + freeze-last (F2)** is exactly right across 28→29→30: B empty → Veto-Governor (constructor reads `B.getMinDelay()`) → grants → revoke self-admin → renounce deployer. The revoke-before-renounce ordering is correct (reversed, the self-admin would be stranded), and deploy_30's `deployerHasAdmin` precondition makes the script safely re-runnable after a partial failure.
- **`contracts/Timelock.sol` wrapper** (OZ 4.8.3, 3-param, passes `msg.sender` as the 4th `admin` arg) matches the "admin = deployer" intent; empty `proposers[]`/`executors[]` is the documented iterative-wiring variant.
- **deploy_29 constructor arg order** matches `GovernorOLAS.sol` exactly (token, timelock, votingDelay, votingPeriod, threshold, quorumFraction, governorDelay).
- **globals_mainnet.json values** all match the spec: `0 / 0 / 13091 / 19636 / 5000e18 / 3`; `governorAddress` = current `0x060D…51E6`, `wveOLASAddress` = `0x4039…4B40` (both re-verified on-chain).
- **The F1 hard gate** — `veto.token() == mainGovernor.token()` with abort — is the executable deploy artifact the #120 review asked for. 
- **Task-2 echo calldata** correct (`grantRole(CANCELLER, VT)`, `setVotingDelay(72000)`, `updateGovernorDelay(1209600)`).

### BLOCKER — resolve before anyone runs deploy_30 (the freeze is irreversible)

**deploy_30 + README intentionally do NOT grant `CANCELLER_ROLE` to the Veto-Governor on VT — and the post-freeze gate *fails* if it is granted (`cancOK` must be `false`). This directly contradicts the merged rollout doc**, whose Task 1 Step 3 grants `PROPOSER + CANCELLER + EXECUTOR` and whose post-deploy check requires `B.hasRole(CANCELLER_ROLE, VetoGov) = true` — and it contradicts your own comment on #120 (which announced #212 grants all three; the commit, pushed later that day, reversed this without a note). An operator following both documents ends in a guaranteed verification failure — and whichever way it lands is **permanent** post-freeze.

Two further points on the stated rationale:

1. *"reachable only in the zero-second window between queue() and execute()"* — not accurate. The **delay** is zero but the **window is unbounded**: `queue()` and `execute()` are separate permissionless transactions, and a succeeded-and-queued veto proposal sits pending on VT indefinitely until someone executes it. During that whole window, `GovernorCompatibilityBravo.cancel(uint256)` (proposer self-withdrawal, or anyone once the proposer drops below 5,000) calls `_timelock.cancel` and **reverts forever** without the grant.
2. Related observation either way (also for the Task-3 monitoring list): a queued-but-unexecuted hostile VT op (e.g. a passed `VT.updateDelay(1)` disarm held as a standing kill-switch) is cancellable by *nobody* in both variants (Bravo cancel is proposer-gated; a veto proposal executing `VT.cancel` fails since VT holds no CANCELLER on itself). The defenders' countermeasure is to **execute it themselves at a harmless time** and immediately run the B′ recovery — which needs an alert on VT `CallScheduled`. Worth one line in the ops docs.

**Recommendation:** add the `CANCELLER` grant (step between 1 and 2, flip the gate to `cancOK == true`). It costs nothing, is frozen immutably either way, restores the only functionality the omission kills (proposer withdrawal / below-threshold cleanup of queued veto proposals), and re-aligns with the spec-of-record that three reviewers signed off on. If you feel strongly about minimalism, the alternative is amending the merged rollout doc + audit addendum in autonolas-analytics *first* — but the "zero-second window" justification should be corrected regardless.

### Should-fix

1. **deploy_29's votingDelay guard inverts after Task 2 executes.** It aborts unless `globals.vetoVotingDelay == mainGovernor.votingDelay()` (live). Pre-rollout that's a fine tripwire — but in the documented **B′-redeploy recovery** (T9 disarm), live main `votingDelay` will be **72000**, and the abort message ("fix globals.vetoVotingDelay and re-run") instructs the operator to copy exactly the value the design forbids (veto cycle ~12.7 d ≈ overflows the 14 d window). Replace with the invariant that stays true forever: `(vetoVotingDelay + vetoVotingPeriod) × 12 s + margin ≤ mainGovernor.governorDelay()` — or pin the expected constants — and drop the "match live main" instruction.
2. **Fork-test drift vs the analytics suite (fixed there today, `7083f7c`):** `D_A` is still `10 days`; the rollout ships **14 days (`1209600`)** — sync it (the analytics copy re-verified 12/12 after the same change). Note the suite's `assertFalse(hasRole(CANCELLER…))` at setUp ties the tests to this PR's side of the blocker — flip it together with whatever deploy_30 lands on.
3. **The test can't run from a fresh clone:** `forge-std` is not in `lib/` (submodules: solmate, fx-portal only), not in `package.json`, and no README note or CI job covers it. Add the submodule (or a setup line in the deployment README) — and ideally a CI job, since this suite is the regression net for the whole veto stack.
4. **Real-deploy fork test (cheap, closes the stand-in gap):** this repo *has* `GovernorOLAS` — one additional test that actually does `new GovernorOLAS(wveOLAS, B, 13091, 19636, 5000e18, 3, 0)` on the fork and asserts `token() == mainGovernor.token()` would turn the F1/F2 doc guards into a CI-checked artifact, complementing deploy_29's cast-based gate.

### Nits

- deploy_30 header says "After step 5" — there are 4 steps.
- Failure paths `exit 0` (repo convention, but automation will read success) — at least the post-freeze verification failure correctly `exit 1`s.
- setUp pins `governorDelay == 157092` / `minDelay == 0` — correct today, will need updating post-Task-2 (expected; noting so it doesn't surprise).

### Verdict

Scripts are structurally right and the order constraints from #120's F2 are faithfully encoded — good work. **Not merge-ready until the CANCELLER contradiction is resolved one way or the other** (it's a one-line change here or a doc amendment there, but it must be decided before the irreversible freeze ever runs). With that + the votingDelay guard fix, this is the Task-1 artifact the rollout doc now points to.

scripts/deployment/deploy_30_wire_freeze_veto_timelock.sh
```diff
@@ -0,0 +1,153 @@
...
+  echo "${red}!!! deployer does NOT hold TIMELOCK_ADMIN_ROLE on Veto Timelock.${reset}"
+  echo "${red}    The freeze already ran or the deployer changed. Abort.${reset}"
+  exit 0
+fi
+
+send() {
+  local step="$1"; local sig="$2"; local addr="$3"; shift 3
+  echo "${yellow}[$step] cast send $addr $sig $@${reset}"
+  cast send --rpc-url $rpcURL $walletArgs $addr "$sig" "$@" >/dev/null
+}
```
dagacha, line 118: [P1] `send()` swallows failed transactions — the freeze can execute in the same run as a failed grant

`cast send` output is piped to `/dev/null` with no exit-status check, and the script has no `set -e`. If step 1 or 2 fails (ledger reject, RPC error, reverted tx), execution falls through to steps 3–4 — the irreversible revoke/renounce — in the same run, leaving a VT with no admin, no proposer, no executor. The post-freeze verification catches it, but only after the renounce has landed. This also qualifies the "safely re-runnable after a partial failure" property: the `deployerHasAdmin` precondition only helps if the script actually stops at the failure.

```suggestion
send() {
  local step="$1"; local sig="$2"; local addr="$3"; shift 3
  echo "${yellow}[$step] cast send $addr $sig $@${reset}"
  cast send --rpc-url $rpcURL $walletArgs $addr "$sig" "$@" >/dev/null || {
    echo "${red}!!! [$step] transaction failed — aborting BEFORE the freeze.${reset}"
    exit 1
  }
}
```

test/forge/ForkGovernanceVetoDelay.t.sol
```diff
@@ -0,0 +1,737 @@
...
+        A.execute(TIMELOCK_A, 0, badData, NO_PREDECESSOR, bytes32(uint256(0xBAD)));
+        assertTrue(A.hasRole(CANCELLER_ROLE, ATTACKER), "bad op fired: ATTACKER gained CANCELLER");
+        assertTrue(A.isOperationDone(badId), "bad op is now Done");
+
+        // Warp to good's eta and try to execute the cancel - reverts because badId is no longer
+        // Pending on A (isOperationPending is false for a Done op). The trap has closed.
+        vm.warp(goodEta);
+        vm.prank(GOVERNOR_A);
+        vm.expectRevert(bytes("TimelockController: underlying transaction reverted"));
+        A.execute(TIMELOCK_A, 0, goodData, NO_PREDECESSOR, goodSalt);
```
dagacha, line 334: [P1] T2's revert assertion passes for the wrong reason

The comment says the execute reverts "because badId is no longer Pending". In reality the inner self-call `A.cancel(badId)` runs with `msg.sender = TIMELOCK_A`, and OZ 4.8.3 `cancel` is `onlyRole(CANCELLER_ROLE)` — checked *before* the `isOperationPending` require. Timelock A does not hold CANCELLER on itself on live mainnet, so this reverts with the identical bubbled `"TimelockController: underlying transaction reverted"` even if the bad op had never executed. The `expectRevert` passes without demonstrating the race the test exists to pin. Fix: grant CANCELLER to `TIMELOCK_A` in the test first (`vm.prank(TIMELOCK_A); A.grantRole(CANCELLER_ROLE, TIMELOCK_A);`) so the revert genuinely comes from the not-Pending check, or assert `isOperationDone(badId) && !isOperationPending(badId)` directly and reframe the execute-revert as incidental.

test/forge/ForkGovernanceVetoDelay.t.sol
```diff
@@ -0,0 +1,737 @@
...
+        // "underlying transaction reverted" (OZ TimelockController._execute:354). The Veto-Governor
+        // proposal simply fails; because Governor.execute is retryable and the B op does not expire
+        // (queued at delay 0 with no expiry), the veto retries once the attacker actually queues.
+        vm.prank(VETO_GOVERNOR);
+        vm.expectRevert(bytes("TimelockController: underlying transaction reverted"));
+        B.execute(TIMELOCK_A, 0, cancelCalldata, NO_PREDECESSOR, salt);
+
+        // Retry after the bad op appears: prove the same B op is executable again.
+        (bytes32 badId, ) = _schedBadOp_grantRoleToAttacker();
+        assertEq(nonexistent, nonexistent); // silence unused warning; the retry uses a fresh salt
```
dagacha: [P3] No-op assert, and the retryability claim is not actually exercised

`assertEq(nonexistent, nonexistent)` is a tautology, and `nonexistent` is already used above, so there is no warning to silence. More substantively, the docstring claims to prove "the same B op is executable again", but the retry schedules a **new** B op (different calldata, fresh salt) — the original op's retryability is never exercised; what remains is already covered by T3. Either re-execute the original op, or drop the claim and the no-op assert.

dagacha (commented):
Independently reviewed the scripts and fork suite; three inline findings below (one P1 on deploy_30's failure handling, one P1 + one P3 on the fork tests) — none previously raised.

One nuance on the CANCELLER blocker discussion: whichever way it resolves, granting CANCELLER to the Veto-Governor on VT does **not** address the queued-but-unexecuted-hostile-VT-op residual — Bravo `cancel(uint256)` is proposer-gated, so the grant only restores proposer self-withdrawal / below-threshold cleanup. The grant decision and the ops-monitoring line for standing VT ops are orthogonal and can be settled independently.

kupermind:
Thanks @DavidMinarsch and @dagacha — both reviews landed and are now addressed in `2f4ac67`. 12/12 fork tests still pass on live mainnet fork.

### BLOCKER — CANCELLER on VT: granted

You're right — I was wrong on the "zero-second window" framing. `queue()` and `execute()` are separate permissionless txs, so the queued-but-unexecuted window is unbounded. During that window `GovernorCompatibilityBravo.cancel(uint256)` (proposer self-withdrawal, or anyone once proposer drops below 5,000 veOLAS) reaches `_timelock.cancel` and would revert forever without the grant. That's a functional regression, not just aesthetic.

Fix in `deploy_30`:
- Sequence is now 5 steps: `PROPOSER → EXECUTOR → CANCELLER → revoke self-admin → renounce deployer`.
- Post-freeze gate flipped: `cancOK` must be `true`.
- Header comment corrected — no more "zero-second window" claim.

I've also flagged @dagacha's second observation in the ops docs (a queued-but-unexecuted hostile VT op is cancellable by *nobody* under either variant, so the CANCELLER grant doesn't help there — the defenders' countermeasure is to execute it at a harmless time and B'-recover). That's an alerting/runbook item, not a code change here; I'll add the `VT CallScheduled` alert requirement to the ops list.

### dagacha P1 — deploy_30 `send()` swallowed failures

Fixed. `send()` now checks `cast send`'s exit code and hard-aborts before any subsequent step runs, so a failed grant cannot fall through to the irreversible revoke/renounce. Message points the operator at the safe re-run path (deployer still holds `TIMELOCK_ADMIN_ROLE` until step 5).

### David should-fix — `deploy_29` votingDelay guard inverts post-Task-2

Fixed. Replaced the "match live main `votingDelay`" check with the stable design invariant:

```
(vetoVotingDelay + vetoVotingPeriod) * 12 s + margin ≤ 1209600 s  (14 d, Task-2 target)
```

Plus a pin (`vetoVotingDelay == 13091 && vetoVotingPeriod == 19636`) as a typo-catch. This is correct in every state — pre-Task-2, post-Task-2, and during B'-recovery (when live main `votingDelay = 72000`, which the check would otherwise ask the operator to copy).

### David should-fix — fork-test D_A drift 10 → 14 days

Fixed. `D_A = 14 days` now matches the rollout constant. The Layer-1 `votingDelay = 72000` blocks (~10 days) references are unchanged — that's the separate Layer-1 lever, not `governorDelay`.

### David should-fix — `forge-std` not in `lib/`

Added a "Running the fork tests" subsection to `scripts/deployment/README.md` with:
```
forge install foundry-rs/forge-std --no-commit
ETH_RPC_URL=<rpc> forge test --match-contract ForkGovernanceVetoDelay -vv
```
Kept as a documented `forge install` line rather than adding a submodule to avoid a repo-wide layout change; happy to convert to a submodule if preferred.

Real-deploy fork test (David's other should-fix) — deferred as a follow-up; the cast-based F1 gate in `deploy_29` already surfaces the token-mismatch loudly at deploy time. Happy to add if you want it as a CI-checked artifact.

### dagacha P1 — T2 assertion passed for wrong reason

Fixed. `A.grantRole(CANCELLER_ROLE, TIMELOCK_A)` before the retry attempt, so the revert genuinely comes from the `isOperationPending == false` trap (not from `onlyRole(CANCELLER_ROLE)`). Added an `assertFalse(isOperationPending)` sanity assertion + comment explaining the ordering of OZ's checks. Same bubbled string, but now for the right reason.

### dagacha P3 — precondition test no-op assert + retryability claim

Fixed. Dropped the `assertEq(nonexistent, nonexistent)` tautology; corrected the docstring to describe the actual operational retry pattern (a fresh Veto-Governor proposal targeting the observed `badId`, not re-executing the original op with a nonexistent id). The retry-with-fresh-salt block still exercises the "veto reverts atomically, no state written on A, retry succeeds" path — which is the property that matters.

### Nits

- "After step 5" — self-resolves with the CANCELLER addition (there are now 5 steps).
- `exit 0` on failure paths — kept to match repo convention on the other deploy scripts; the load-bearing post-freeze verification correctly `exit 1`s on any deviation.
- setUp assertions `governorDelay == 157092` / `minDelay == 0` — noted; these will need updating post-Task-2 execution.

### Summary of what changed

- `scripts/deployment/deploy_29_veto_governor.sh` — new invariant check, drop live-match.
- `scripts/deployment/deploy_30_wire_freeze_veto_timelock.sh` — CANCELLER grant, hardened `send()`, renumbered steps.
- `scripts/deployment/README.md` — CANCELLER grant text, forge-std setup, invariant description.
- `test/forge/ForkGovernanceVetoDelay.t.sol` — CANCELLER in setUp, `D_A = 14 days`, T2 fix, precondition-test cleanup.

Suite: **12/12 PASS** on live mainnet fork.

DavidMinarsch:
## Re-review of `2f4ac67` — all findings verified fixed; merge-ready

Independently verified every fix in the diff and **re-ran the suite at `2f4ac67` on a live mainnet fork: 12/12 PASS** (T2's changed gas confirms the fixed assertion path is the one running).

- **BLOCKER (CANCELLER on VT): resolved correctly.** Grant added as step 3/5, post-freeze gate flipped to `cancOK == true`, header and README rationale corrected (no more "zero-second window"). Now exactly matches the merged rollout doc's Task 1 Step 3 + post-deploy checks. The "VT can only cancel proposals it itself queued" note is accurate given VetoGov is VT's sole proposer.
- **`send()` hard-abort (dagacha P1): correct.** Exit-code check with `exit 1` *before* any subsequent step — a failed grant can no longer fall through to the irreversible revoke/renounce, and the message correctly points at the safe re-run path (deployer admin intact until step 5).
- **deploy_29 invariant: correct in every state.** Pinned constants (13091/19636) + `(vD + vP) × 12 s + 86400 ≤ 1209600` — arithmetic checks out (392,724 + 86,400 = 479,124 s ≪ 14 d), and checking against the Task-2 design constant rather than live state is right twice over: a live-`governorDelay` check would also spuriously fail *pre*-Task-2 (live 157092 s < veto cycle), and a live-`votingDelay` match would poison B′-recovery. Good fix.
- **T2 wrong-reason fix (dagacha P1): correct.** Granting CANCELLER to A-on-itself (prank via self-admin) plus the `assertFalse(isOperationPending)` sanity makes the bubbled revert genuinely prove the pending-trap. Side observation the comment rightly implies: in production the same-Timelock cancel route is *doubly* dead (A holds no CANCELLER on itself, and the race), which only strengthens the case for the separate VT.
- **Precondition test (dagacha P3):** tautology gone, docstring now describes the real fresh-proposal retry pattern. ✓
- **`D_A = 14 days`**, T3 docstring at 9.45 d margin, README forge-std section (the `ForkDeployGovernance.t.sol` reference is valid — it exists) — all ✓.
- **Deferred real-deploy fork test:** accepted as a follow-up; deploy_29's cast-based F1 gate covers it operationally.

One optional one-liner, non-blocking: the analytics repo's merged copy of this suite (autonolas-analytics `forge/test/ForkGovernanceVetoDelay.t.sol`) wires VT in setUp with PROPOSER+EXECUTOR only — nothing there exercises Bravo cancel so nothing is wrong, but a future sync adding the CANCELLER grant would keep the two suites' setUp identical to deploy_30's five steps.

**Verdict: merge-ready.** With this merged, the Task-1 deploy artifact the rollout doc points to is real, gated, and fork-verified. Remaining sequence: relock (2026-07-09 — now 3 days), Task-1 deploys, batched Task-2 vote, Task-3 UI/alerts (incl. `VT.getMinDelay() != 0` and the VT `CallScheduled` alert from the review thread).

DavidMinarsch (approved):

dagacha:
Verified `2f4ac67` against my three inline findings — all fixed:

- **P1 — deploy_30 `send()`**: now hard-aborts (`exit 1`) on a failed `cast send` before any subsequent step, so a failed grant can no longer fall through to the irreversible revoke/renounce. ✅
- **P1 — T2 wrong-reason revert**: CANCELLER granted to A-on-itself via self-admin prank + `assertFalse(isOperationPending(badId))` sanity — the bubbled revert now genuinely proves the pending-trap. ✅
- **P3 — precondition test**: tautological assert removed; docstring now matches what the retry block actually exercises (fresh veto op targeting the observed `badId`). ✅

Also spot-checked the deploy_29 invariant arithmetic — `(13091 + 19636) × 12 + 86400 = 479,124 s ≪ 1,209,600 s` — and that the CANCELLER resolution landed consistently across deploy_30, the post-freeze gate, README, and the test setUp.

No open findings from my side — concur with @DavidMinarsch: merge-ready.

dagacha (approved):

DavidMinarsch:
## Proposed revision: `VetoTimelock` — close the T9 self-disarm by construction

@kupermind — one design revision to fold in before this merges, and it reverses a position I stated on the #120 thread ("on-chain prevention of T9 was considered and declined"), so laying out the reasoning openly.

### The change

Deploy Timelock B not as stock `Timelock` but as a minimal subclass:

```solidity
// contracts/VetoTimelock.sol
contract VetoTimelock is TimelockController {
    constructor(uint256 minDelay, address[] memory proposers, address[] memory executors)
        TimelockController(minDelay, proposers, executors, msg.sender) {}

    /// @dev minDelay is frozen at its construction value forever (veto cancels must stay instant).
    function updateDelay(uint256) external pure override {
        revert("VetoTimelock: minDelay frozen");
    }
}
```

`updateDelay` is the one privileged timelock function outside the role system (`onlySelf`, role-less) — the only self-modification path the role-freeze cannot reach, and the entire basis of the T9 residual (one won veto vote → `B.updateDelay(1)` → Layer 2 dark for a full B′-redeploy + re-grant cycle, ~27 days post-rollout). With the override, that call reverts; after the freeze **nothing about B can change at all** — roles frozen, delay frozen, cancel-only by construction with zero self-modification surface.

### Why I reversed

My "declined" rested on the stock-bytecode premise — the same ground §11.3 rejected the target-allowlist and §11.5 the one-contract variant. But those were **classification logic** (code that decides, and can be wrong in both directions). A pure-revert override is a **monotone restriction**: it removes a capability and cannot add, decide, or misclassify anything. It is the same move as the §3 role-freeze — converting a declared invariant into a constructed one — applied to the last mutable knob. Two more points that finished it:

- "Zero new Solidity" was already only approximately true: `contracts/Timelock.sol` is itself a subclass (constructor wrapper). `VetoTimelock` is that wrapper plus one revert — the delta from existing practice is one function.
- What it deletes is bigger than the attack itself: the recovery runbook, the ~27-day Layer-2-dark window, the queued-disarm-as-timed-kill-switch scenario, **and the go-live monitoring** — a `getMinDelay` alarm watching a value that can no longer change is noise, so the Task-3 alert list shrinks to the propose-time shape check only.

Mechanics verified against the deployed lineage: `updateDelay` is `virtual`; the constructor writes `_minDelay` directly (construction unaffected); nothing else calls it internally; the Veto-Governor constructor only *reads* `getMinDelay()`; and it is the only role-less self-call state mutator on the contract, so one override closes the whole class.

### Concrete diffs this implies

1. **`contracts/VetoTimelock.sol`** — as above (naming yours to bikeshed).
2. **`deploy_28`** — `contractName="VetoTimelock"`; everything else unchanged (same 3-arg shape, `msg.sender` admin).
3. **Fork suite** — `setUp` deploys `VetoTimelock`; `test_T9` inverts: the veto proposal executing `B.updateDelay(1)` must **revert** (`expectRevert("VetoTimelock: minDelay frozen")` bubbled as "underlying transaction reverted"), `getMinDelay()` stays 0, and a subsequent veto cancel through B still works — Layer 2 never goes dark.
4. **README** — drop the `getMinDelay != 0` monitoring line; note B is fully immutable post-freeze.

### Residual honesty

This does **not** touch the griefing residual (an attacker who can win veto votes can still grief-cancel — same bar, inherent to any cancel brake, bounded by T8b removability). It eliminates the cheaper, stealthier member of that threat class, not the class.

The AIP-8 draft has been updated to assume this revision (review PR: valory-xyz/autonolas-analytics#124 — you're a reviewer). If you'd rather keep stock and argue the other side, say so there — but the freeze decision is permanent either way, so it must land before anyone runs `deploy_30`.

kupermind:
Concur — this is the right final form. Verified the override mechanics against OZ v4.8.3 `TimelockController.sol`:

- `updateDelay` is `virtual` (L383) — the override compiles.
- The constructor writes `_minDelay` directly (L107), not via `updateDelay` — construction still works.
- `updateDelay` is the only state-changing function outside the role system on `TimelockController`. Every other write goes through PROPOSER / EXECUTOR / CANCELLER / role admin — all frozen at §3. One override closes the last remaining self-modification path on B.
- The Veto-Governor only *reads* `getMinDelay()` (for the `governorDelay >= minDelay` check) — the write-side revert doesn't touch it.

Result: after the freeze nothing on B can change, T9 is closed by construction, and the `getMinDelay` alarm drops out of the runbook. Agreed.

### One naming/subclassing nit

Prefer `is Timelock` (the existing thin wrapper) over `is TimelockController` in the snippet. `contracts/Timelock.sol` already delegates `msg.sender` as admin, so:

```solidity
contract VetoTimelock is Timelock {
    constructor(uint256 minDelay, address[] memory proposers, address[] memory executors)
        Timelock(minDelay, proposers, executors)
    {}

    function updateDelay(uint256) external pure override {
        revert("VetoTimelock: minDelay frozen");
    }
}
```

Same bytecode, one function delta from something we already ship. `external pure` is the tightest correct signature. Bikeshed accepted either way.

### Follow-up PR I can open

Happy to pick up the code deltas once the AIP text settles and dagacha weighs in:

1. **`contracts/VetoTimelock.sol`** — as above.
2. **`scripts/deployment/deploy_28_veto_timelock.sh`** — `contractName="VetoTimelock"`; ctor args unchanged.
3. **`test/forge/ForkGovernanceVetoDelay.t.sol`** — `setUp` deploys `VetoTimelock`; `test_T9` inverts to assert the disarm reverts + `getMinDelay()` stays 0 + a subsequent veto cancel through B still succeeds. Rename to something like `test_T9_veto_stack_self_disarm_reverts_and_layer2_stays_live`.
4. **`scripts/deployment/README.md`** — drop the `getMinDelay != 0` monitoring line; add "post-freeze, no state on B can change — no runtime monitoring is required on VT".

The CANCELLER grant + hardened `send()` + design-invariant `deploy_29` guard from #212's review-fix commit are orthogonal and stay as-is.

Ready to move once dagacha's in and the AIP text is final.

DavidMinarsch:
Converged — thanks for the independent mechanics verification (matches mine point for point), and your `is Timelock` wrapper-subclass form is the implementation reference: same one-function delta from something already shipped, `external pure` as the tightest signature. Bikeshed settled in your favour.

Status on your two wait-conditions:

- **AIP text:** settled modulo dagacha — the draft (analytics #124) is updated to assume `VetoTimelock` (Specification, Security Considerations "self-disarm closed by construction", T9 inversion in Test Cases, no B monitoring anywhere), and your review notes there are all addressed in `be5719f`. I've approved your four code deltas as listed, including the `test_T9_…_reverts_and_layer2_stays_live` rename — the "subsequent veto cancel through B still succeeds" post-condition is the load-bearing assertion, since it proves Layer 2 is genuinely undisarmable rather than merely revert-y.

- **One sequencing request: fold the deltas into THIS PR rather than a follow-up.** If #212 merges as-is, `main` carries a stock-`Timelock` `deploy_28` that must never actually be run — a deploy artifact whose correction lives in an unmerged second PR, guarding an irreversible freeze. That's the same shape of landmine as F2, just at the repo level. The four deltas touch only files already in this PR's diff (`deploy_28`, the fork suite, the deployment README) plus one new 8-line contract, so the revision is small, and the PR then merges as the single self-consistent Task-1 artifact the AIP references. If you'd rather keep the follow-up-PR shape for review hygiene, the alternative is a loud `!!! DO NOT RUN — superseded by VetoTimelock (PR #XXX)` guard at the top of `deploy_28` in this PR — but in-place is cleaner.

No rush on the push itself — dagacha's #124 review may still adjust AIP wording, though the mechanics are triple-verified at this point and won't move.

dagacha:
Replying to @DavidMinarsch's latest `VetoTimelock` comment, specifically the claim that the “subsequent veto cancel through B still succeeds” post-condition proves Layer 2 is genuinely undisarmable rather than merely revert-y:

Agree that the “subsequent veto cancel through B still succeeds” post-condition is the right assertion for proving `updateDelay` no longer bricks Layer 2.

But I’d qualify the conclusion: this makes Layer 2 undisarmable **via `B.updateDelay`**, not undisarmable in general. Two same-cost disarm paths appear to remain, both requiring exactly what T9 required: one won veto vote. Both then execute immediately at `vetoGovernorDelay = 0` / `B.minDelay = 0`, with no post-passage counter-veto window.

1. **Veto-Governor self-mutators.**

   A winning veto proposal executes through B, so the `_executor()` / `onlyGovernance` path is satisfied for self-targeted Veto-Governor calls. That means a veto proposal can still call inherited setters such as:

   - `updateTimelock(...)`
   - `updateGovernorDelay(...)`
   - `setVotingDelay(...)`
   - `setVotingPeriod(...)`
   - `setProposalThreshold(...)`
   - `updateQuorumNumerator(...)`

   For example, `updateTimelock(...)` can disconnect the Veto-Governor from the B that holds `CANCELLER_ROLE` on Timelock A, and an absurd `proposalThreshold` can brick future veto proposals.

2. **Role-renounce via `relay`.**

   `renounceRole` is not admin-gated, only `account == msg.sender`-gated. A veto proposal can call:

   ```solidity
   VetoGovernor.relay(
       address(B),
       0,
       abi.encodeCall(
           AccessControl.renounceRole,
           (PROPOSER_ROLE, address(VetoGovernor))
       )
   );
   ```

   `relay` executes with the Veto-Governor as `msg.sender`, so the Governor renounces its own `PROPOSER_ROLE` on B. The same works for `EXECUTOR_ROLE`; `CANCELLER_ROLE` is less fatal but still degrades Bravo cleanup. Post-freeze B has no `TIMELOCK_ADMIN_ROLE` holder, so the loss is unrecoverable on B itself: full B′ / Veto-Governor redeploy plus one main-Governor cycle to re-grant `CANCELLER_ROLE` on A, the same ~27-day Layer-2-dark window T9 described.

So `VetoTimelock` reduces the one-won-vote disarm paths from N to N−1; it doesn’t close the class by construction.

Concrete asks:

- **AIP / README wording:** say “`VetoTimelock` freezes B’s `minDelay` and closes the `B.updateDelay` self-disarm vector,” not “self-disarm closed by construction” / “post-freeze, no state on B can change.”
- **Monitoring:** drop the `getMinDelay` alarm, which is genuinely dead, but keep runtime alerts on B `RoleRevoked` and Veto-Governor `TimelockChange` + settings-change events (`VotingDelaySet`, `VotingPeriodSet`, `ProposalThresholdSet`, `QuorumNumeratorUpdated`, `GovernorDelayChange`).
- **Fork tests:** alongside the inverted T9, add a T9b that either demonstrates the `relay`-renounce path on the fork or records it explicitly as an accepted residual.
- **Optional follow-up hardening:** the same monotone-restriction argument that justified `VetoTimelock` applies Governor-side, a `VetoGovernorOLAS` with reverting overrides for the self-mutators and `relay` would close the class. Same small-delta-from-shipped-bytecode shape.

None of this blocks folding the `VetoTimelock` deltas into this PR, which I agree is the right sequencing. It’s about the AIP and README not claiming a stronger property than the code provides.

Assessment note: this assessment was initiated by GPT-5.5 and reworked with Fable 5.

kupermind:
@dagacha — you're right, and my "closed by construction" line was too strong. Verified both paths against OZ v4.8.3:

- **Path 1 (self-mutators):** `GovernorTimelockControl.updateTimelock` at `GovernorTimelockControl.sol:158` and every settings function on `Governor` are `external virtual onlyGovernance`. Because Veto-Governor's `_executor()` returns B, any veto proposal that targets Veto-Governor itself arrives with `msg.sender == B == _executor()` and passes `onlyGovernance` after the `_governanceCall` deque check (Governor.sol:65–70, filled by `_beforeExecute` at :340–343). So `updateTimelock` (disconnect Veto-Governor from B), `updateGovernorDelay`, `setVotingDelay/Period/ProposalThreshold`, `updateQuorumNumerator` are all reachable via a single won veto vote.
- **Path 2 (`relay` → `renounceRole`):** `Governor.relay(target, value, data) external payable virtual onlyGovernance` at Governor.sol:543 calls `target.call(data)` with `msg.sender = VetoGov`. `AccessControl.renounceRole` is `account == msg.sender`-gated, so `VetoGov.relay(B, 0, encodeCall(renounceRole, (PROPOSER_ROLE, VetoGov)))` self-renounces. Post-freeze B is unrecoverable — same ~27-day Layer-2-dark window T9 originally described.

OZ's own comment at `GovernorTimelockControl.sol:20` flags this class: *"{onlyGovernance} functions like {relay} are…"*. The `VetoTimelock` override reduces the disarm surface from N to N−1, not to zero.

### My preference: close the class the same way — `VetoGovernorOLAS`

The monotone-restriction argument that justifies overriding `B.updateDelay` applies identically Governor-side: revert-only overrides remove capabilities without adding logic that could misclassify. The Veto-Governor is already "redeploy to change" by design, so overriding its self-mutators + `relay` to revert doesn't block any legitimate flow. Stopping halfway leaves an accepted residual at the same one-vote cost as the vector we just closed, plus reinstates the runtime alarms we're trying to retire.

Proposed shape (subclass of `GovernorOLAS`, all overrides `external pure`, all reverting with a stable string per function):

```solidity
contract VetoGovernorOLAS is GovernorOLAS {
    constructor(...) GovernorOLAS(...) {}

    function relay(address, uint256, bytes calldata) external payable override { revert("VetoGovernorOLAS: relay disabled"); }
    function updateTimelock(TimelockController)          external override { revert("VetoGovernorOLAS: timelock frozen"); }
    function updateGovernorDelay(uint256)                external override { revert("VetoGovernorOLAS: governorDelay frozen"); }
    function setVotingDelay(uint256)                     external override { revert("VetoGovernorOLAS: votingDelay frozen"); }
    function setVotingPeriod(uint256)                    external override { revert("VetoGovernorOLAS: votingPeriod frozen"); }
    function setProposalThreshold(uint256)               external override { revert("VetoGovernorOLAS: proposalThreshold frozen"); }
    function updateQuorumNumerator(uint256)              external override { revert("VetoGovernorOLAS: quorum frozen"); }
}
```

`relay` is the only one that keeps `payable` (matches the base sig). All others are non-`payable` on base. `external pure` on most is correct — none read state; `payable relay` can't be `pure`.

Verified the base signatures compile as overrides against the deployed `GovernorOLAS` inheritance chain — the seven functions above are exactly the `onlyGovernance` mutators reachable from a passed veto proposal. (`_updateGovernorDelay` is `internal` and only reached via `updateGovernorDelay` — closed transitively. Quorum via `_updateQuorumNumerator` same story.)

### Concrete follow-up-PR scope

The #212-merge state means this ships as a new PR against `main`:

1. `contracts/VetoTimelock.sol` — as agreed (`is Timelock`, override `updateDelay`).
2. `contracts/VetoGovernorOLAS.sol` — as above.
3. `scripts/deployment/deploy_28_veto_timelock.sh` — `contractName="VetoTimelock"`.
4. `scripts/deployment/deploy_29_veto_governor.sh` — `contractName="VetoGovernorOLAS"`; ctor args + F1 token cross-check unchanged (still reads `token()` off the deployed instance).
5. `test/forge/ForkGovernanceVetoDelay.t.sol`:
   - `setUp` deploys `VetoTimelock` + `VetoGovernorOLAS`.
   - `test_T9` inverts to assert `B.updateDelay(1)` reverts + `getMinDelay()` stays 0 + subsequent veto cancel through B still succeeds. Renamed to `test_T9_B_updateDelay_reverts_and_layer2_stays_live`.
   - **New `test_T9b_governor_self_mutators_revert_and_layer2_stays_live`** covering both surviving paths:
     (a) veto proposal targeting `VetoGov.updateTimelock(newTimelock)` must revert (bubbled), Veto-Governor still bound to B, subsequent veto cancel still works.
     (b) veto proposal targeting `VetoGov.relay(B, 0, renounceRole(PROPOSER_ROLE, VetoGov))` must revert (bubbled), Veto-Governor still has PROPOSER on B, subsequent veto cancel still works.
6. `scripts/deployment/README.md` — B is fully immutable post-freeze; drop `getMinDelay != 0` alarm; drop Veto-Governor `TimelockChange` + settings-change alerts; only propose-time shape check on veto proposals is retained (proposal hygiene).

### On David's other threads

- **Sequencing (fold into #212):** #212 is already merged, so this is a new PR against main. No landmine — `deploy_28`'s next writer supersedes the existing one; no operator would run today's `deploy_28` between now and that push. If you want a belt-and-braces guard, one line at the top of the current `deploy_28` on main saying "superseded by VetoTimelock — see #XXX" is trivial and I can add it in a preliminary micro-PR before the main follow-up lands.
- **Griefing / spam analysis (`3a9fba7`):** LGTM. The `max(H, quorum)` cost frame is right, the "persistent griefing = sustained majority = legitimate control" argument is clean, and per-target-op dedup in the UI is the correct spam control at the layer where it belongs.

### AIP-8 wording implied by (B)

If we go with `VetoGovernorOLAS`, the AIP text stays close to its current shape — "self-disarm closed by construction" is genuinely true again, no monitoring on either B or Veto-Governor is required, T9 test cases cover B.updateDelay + T9b covers the Governor-side residual (both revert). The Rationale's monotone-restriction paragraph naturally extends to cover the Governor-side overrides — same argument, applied to the same class of mutators.

If instead we go with **(A) accept residual**, the AIP text needs the walk-back dagacha proposed and monitoring goes back into the runbook. Happy to draft either variant.

@DavidMinarsch — over to you for the call on (A) vs (B). I'm ready to start pushing on the follow-up PR either way; the delta shifts by ~5 files not much more.

77ph:
## Pre-deployment validation of the mainnet run (Task 1: `deploy_28` → `29` → `30`)

Validated the requested procedure against this PR's scripts, the rollout doc (`olas-governance-delay-rollout-2026-07-03.md`), the `Timelock.sol` / `GovernorOLAS.sol` constructors, `globals_mainnet.json`, and **live mainnet state**. Summary: the scripts themselves are sound and the ordering in the request is correct — but **one question must be answered before `deploy_30` runs, because that step is irreversible.**

### 🔴 Blocking question — stock `Timelock` vs the `VetoTimelock` subclass

`deploy_28` deploys **stock `contracts/Timelock.sol`** (`contractName="Timelock"`, and the header says so explicitly), and `deploy_30`'s header records the consequence as accepted:

> `B.updateDelay(...)` is `onlySelf` (role-less), so a veto vote CAN still self-disarm B by bumping its minDelay above 0 (**T9 residual**) — recovery = redeploy B' + re-grant CANCELLER on Timelock A via one main-Governor cycle.

However, the thread above converged **after** this PR merged (2026-07-06) on *not* deploying stock `Timelock`: the `VetoTimelock` minimal subclass overriding `updateDelay` was proposed on 07-07, concurred (“this is the right final form”, with the OZ v4.8.3 override mechanics verified against `TimelockController.sol` L383/L107), and settled as “the implementation reference”. **`contracts/VetoTimelock.sol` does not exist on `main`, and no script references it** — `main` has not moved since 07-07.

So as things stand, running the three scripts today deploys the **pre-decision** design and then **freezes it permanently**: after step 5/5 no account holds `TIMELOCK_ADMIN_ROLE` on B, so the T9 self-disarm cannot be closed later without redeploying B′ *and* re-granting `CANCELLER` on Timelock A through a full main-Governor cycle.

Please confirm explicitly which is intended:
- **(a)** the `VetoTimelock` subclass is still the agreed target → it needs implementing + `deploy_28` re-pointing **before** any mainnet run; or
- **(b)** the team has consciously reverted to stock `Timelock` and re-accepted the T9 residual → then `deploy_30`'s header is the record, and the rollout doc/AIP text should say so, since the thread currently reads the other way.

This is the only item I'd hold the deployment on.

### ✅ Validated as correct

- **Ordering is right, and Task 1 *must* precede Task 2.** The scripts pin `vetoVotingDelay/vetoVotingPeriod = 13091/19636` and abort if changed; the rollout doc requires capturing today's values *before* Task 2 raises `votingDelay` to 72000 (copying the raised value would push the veto cycle to ~12.7 d against a 14 d window). Deploying now is the correct sequencing, not a race.
- **`globals_mainnet.json` is consistent**: `vetoMinDelay = 0`, `vetoGovernorDelay = 0` (instant cancel, and `_updateGovernorDelay(0)` is only valid because `B.minDelay = 0`), `vetoQuorum = 3`, `vetoProposalThreshold = 5000e18`, `wveOLASAddress = 0x4039B809…4B40`. `vetoTimelockAddress` / `vetoGovernorAddress` are absent, as expected — the scripts write them.
- **F1 (token) check is live-verified**: `GovernorOLAS 0x060D0CBd…51E6 .token() = 0x4039B809…4B40` (wveOLAS wrapper) — `deploy_29` cross-checks the veto against this and aborts on mismatch. Correct guard against binding the veto to raw veOLAS `0x7e01A5…`.
- **`deploy_30` fail-safety is right**: `send()` hard-aborts before any subsequent step, so a failed grant can never fall through into the irreversible revoke/renounce; grants (1–3) precede the freeze (4–5); post-state gate asserts `PROPOSER/EXECUTOR/CANCELLER = true` and both `ADMIN` holders `false`.

### ℹ️ State the operators should have in writing

- **The stack is inert until Task 2.** Live `GovernorOLAS.governorDelay() = 157092 s (1.82 d)`, while the veto cycle is `(13091+19636)×12 = 392724 s (4.55 d)`. Until Task 2 raises `governorDelay` to 14 d **and** grants `CANCELLER_ROLE` on Timelock A to B, the veto cannot cancel anything — B holds no privilege on Timelock A at all. That is exactly what the rollout doc specifies (“B's only privilege anywhere is `CANCELLER_ROLE` on Timelock A, granted in Task 2”), so it is by design — but it should be said plainly so nobody reads a successful Task 1 as “the veto is live”.
- **`deploy_29`'s cycle-overflow check is not a readiness gate.** It compares against the pinned design target (`mainGovernorDelayTarget = 1209600`), deliberately *not* live state — the reasoning in the script is sound (a live comparison would misfire during B′-recovery). Worth being explicit that it therefore passes today even though the live 157092 s window does not yet accommodate the veto cycle.
- **Do not release the deployer key between `deploy_28` and `deploy_30`.** `useLedger = true`, and the deployer is the sole `TIMELOCK_ADMIN_ROLE` holder on B for the whole interval; if the run is interrupted after 28/29, B is left admin-held and unwired.

### Minor, on the push-back step

`git commit -am` sweeps every modified tracked file. The intended change is `scripts/deployment/globals_mainnet.json` (the two written addresses); prefer staging it explicitly so an unrelated local edit cannot ride along:

```
git add scripts/deployment/globals_mainnet.json
git commit -m "chore: deployment of veto GovernorOLAS and Timelock"
```

`git clone --recursive` is required (forge-std is a pinned submodule at v1.9.7 per #213), and both `ETHERSCAN_API_KEY` and `ALCHEMY_API_KEY_MAINNET` must be exported — all three scripts abort without them, and `contractVerification = true` means verification runs as part of each step.

kupermind:
Thanks @77ph — this is a thorough pre-flight check and the blocking question is a fair one to raise before anything irreversible runs. Let me answer it clearly, and also tighten up one point so the thread record is accurate.

**First, a clarification on what "VetoTimelock" would and wouldn't be.**

There is no plan for a separately-named `VetoTimelock.sol` in the sense of a rename or a missing artifact. The "Veto Timelock" is simply our existing, audited `contracts/Timelock.sol` deployed a second time with the `veto*` parameters — `deploy_28` deploys `Timelock.sol` with `minDelay = vetoMinDelay (0)` and records the address as `vetoTimelockAddress`. In that sense the scripts are internally consistent and deployable as written; nothing is half-wired.

Where you're right, and where the real question lives, is that closing the T9 self-disarm hole is **not** something a parameter can do. `vetoMinDelay` has to stay `0` for the instant veto to function, so the only way to prevent a passed veto vote from calling `B.updateDelay(>0)` and disabling Layer 2 is a **code override** of `updateDelay` — i.e. a minimal subclass of `Timelock` that makes that function revert. So the choice is genuinely:

- **(a)** add that one-function override (a real, if small, code change: new contract + re-point `deploy_28` + re-run the fork suite), which makes it impossible for a veto vote to ever disarm the veto layer; or
- **(b)** deploy stock `Timelock` and accept the T9 residual — a passed veto vote can raise B's delay and switch Layer 2 off, with the mitigations being (i) it's bounded (the main stack is unaffected — the T9 test proves this), (ii) it takes actually winning a veto vote, not a cheap external action, and (iii) we monitor `B.getMinDelay() != 0` as wired into `deploy_30`.

**This is a design/risk-tolerance decision, and it sits with management — not something we'll settle inside the deploy scripts.** So we're escalating (a) vs (b) to them rather than picking one here. Until they decide, the mainnet run is on hold.

Your irreversibility point is fully taken and is exactly why we're stopping here: after `deploy_30` no account holds `TIMELOCK_ADMIN_ROLE` on B, so if we ship (b) the T9 residual cannot be closed in place later — recovery would mean redeploying a fresh B′ and re-granting `CANCELLER` on Timelock A through a full main-Governor cycle. Not something to discover after the freeze. If management chooses (a), we implement the override and re-verify before any run; if they choose (b), we'll update the rollout doc / AIP text so it plainly records the accepted residual, since as you note the thread currently reads the other way.

**On everything else you validated** — ordering (Task 1 before Task 2, pinned `vetoVotingDelay/Period`), `globals_mainnet.json` values, the live `token()` guard in `deploy_29`, the fail-safe ordering in `deploy_30`, and the "stack is inert until Task 2" note — agreed on all of it, no changes needed. We'll also adopt the explicit `git add scripts/deployment/globals_mainnet.json` instead of `commit -am` so nothing unrelated rides along, and keep the deployer key held across 28→30 as you flag.

Holding the deployment on the (a)/(b) decision. Will update here once management calls it.

PR #215, opened 2026-07-28
fix(VoteWeighting): reconcile removeNominee accounting to close checkpoint DoS

kupermind:
## Summary

Fixes a set of recorded `VoteWeighting` findings on a fresh (non-upgradeable) redeploy of the contract. Headline is **finding #8**: `removeNominee` retired a nominee's bias but left its slope and `changesSum` entries behind, and the unguarded subtraction in `_getSum` could then **permanently revert the weekly checkpoint walk** — which also halts the Dispenser's `nomineeRelativeWeightWrite` path.

Scope decision: **plain non-upgradeable redeploy, security fixes only** — no proxy refactor, no `MAX_NOMINEES` cap (both deferred by design).

## Findings addressed (`docs/Vulnerabilities_list_governance.md`)

- **#8 (primary)** — three coordinated parts:
  1. Guard the aggregate slope subtraction in `_getSum` and `_getWeight` with the contract's own `_maxAndSub`, so the weekly walk can never underflow-revert.
  2. `removeNominee` now **reconciles** the removed nominee out of the aggregate: it subtracts the nominee's still-active slope from `pointsSum`, and strips its future `changesSum` decrements in a bounded loop (matching the `_getSum`/`_getWeight` horizon, which exceeds the 4-year max lock). No phantom slope survives — this is *correct*, not merely non-reverting: other nominees sharing a week are undisturbed.
  3. `revokeRemovedNomineeVotingPower` is reduced to **pure per-user cleanup** (no aggregate writes), since removal now fully reconciles the aggregate — this removes the double-count that would otherwise occur.
- **#11** — `OwnerOnly(msg.sender, owner)` argument order in `removeNominee`.
- **#18** — clamp `_nomineeRelativeWeight` at `1e18` (defends the Dispenser consumer against any residual drift).
- **#19** — last-element swap guard changed from `numNominees > 1` to `id != numNominees`, so removing the last element no longer re-populates `mapNomineeIds` for the just-removed nominee.
- **#20** — resolved by construction: the revoke redesign no longer writes any checkpoint slot, so there is no stale next-week slot to mis-advance.

The same fixes are ported into the `VoteWeightingFuzzing` echidna harness so it fuzzes the corrected logic.

## Testing

- **`test/forge/VoteWeighting.t.sol`** — 17 unit tests over the real OLAS + veOLAS + VoteWeighting stack: the #8 checkpoint-DoS regression (remove-without-unvote past lock expiry, then a previously-underflowing second removal), slope + `changesSum` reconciliation, multi-voter / expired-voter / vote-change / zero-weight-unvote / batch cases, revoke double-count / late-revoke / revoke-twice, last-element and first-element removal, `OwnerOnly` arg order, relative-weight clamp, and a cross-horizon aggregate-consistency invariant (60-week walk). The key regression tests were confirmed to **fail on the pre-fix contract** and pass on the fix.
- **`test/forge/ForkVoteWeighting.t.sol`** — 2 mainnet-fork tests deploying the patched contract against live veOLAS (`0x7e01…`), replaying the remove-without-unvote sequence against real lock math.
- Existing Hardhat suite: **174 passing**, unchanged. `forge build`, `solhint`, and `eslint` clean (no new warnings).

## Test plan

- [x] `forge test --match-contract VoteWeightingTest` — 17/17
- [x] `forge test --fork-url <MAINNET_RPC> --match-contract ForkVoteWeighting` — 2/2
- [x] Regression proof: key tests fail on the pre-fix contract
- [x] `npx hardhat test` — 174 passing
- [x] `solhint` / `eslint` (CI commands) — 0 errors, no new warnings
- [ ] External re-audit of the `removeNominee` / revoke reconciliation before deployment

## Notes

- `removeNominee` gas rises to ~715k (owner-only, rare) due to the bounded reconciliation loop.
- This is a security-critical accounting redesign of `removeNominee` / revoke and should go through the normal re-audit before it ships.
- `docs/Vulnerabilities_list_governance.md` is intentionally left unchanged in this PR.


77ph (dismissed):
**Fix-verification: PASS / merge-ready.**

This closes the `internal20` VoteWeighting "Findings to fix" at the source — F-6 (#8/#11), F-3 (#18), F-4 (#19), F-5 (#20), F-9 (#11) — including the F-6 case for which `internal20` explicitly declined the "ask voters to revoke before removal" operational workaround.

Verified firsthand:
- **`removeNominee` reconciliation is exact.** The slope subtracted from `pointsSum` is read from the same `pointsWeight[N][nextTime].slope` slot the voting path maintains (`voteForNomineeWeights` :543–549), under the same `nextTime` convention, so it's the exact mirror of a vote; the `changesWeight[N][t] → changesSum[t]` strip removes only N's own contribution and doesn't disturb nominees sharing week `t`; `MAX_NUM_WEEKS = 250` covers the maximum 4-year (~209-week) veOLAS lock horizon.
- **`revoke` no longer double-subtracts.** With the aggregate fully reconciled in `removeNominee`, reducing `revoke` to pure per-user cleanup is correct — and makes the aggregate correct regardless of whether users ever revoke (the fatal dependency of the previous design). Removal is terminal (`_addNominee` :304 / `voteForNomineeWeights` :487), so no stale per-user slope can be reactivated.
- **Tests bind the bug.** `forge test --mc VoteWeighting` → 17/17 unit pass on the fix. Reverting only the contract to the pre-fix parent (`a7f6189`) while keeping the new tests, the three core regressions (`NoCheckpointDoS`, `ReconcilesSlopeAndChangesSum`, `MultiVoter_FullReconcile`) fail on their root-invariant assertions.

Non-blocking: the `MAX_NUM_WEEKS` strip loop in `removeNominee` is owner-gated (governance → Timelock), so its worst-case gas has no griefing surface; and on this fresh non-upgradeable redeploy the `_maxAndSub` guards are defense in depth.

Full write-up in the fix-verification report: #216 (`audits/internal21`).

77ph (dismissed):
## Re-review after the post-approval delta (my earlier approval was auto-dismissed by the new commits)

Re-verified the delta `a22490e … 8326d365` (4 commits since the first approval). **Still merge-ready.**

**The removeNominee #8 reconciliation is logically unchanged.** The edits to `_getSum` / `_getWeight` / `removeNominee` / `revoke` are comment-text only (the `(finding #N)` tags were dropped); the reconciliation code — `nomineeSlope` capture, `_maxAndSub` guards, the `changesSum` strip loop, and the pure-per-user `revoke` — is byte-identical to what was verified.

**New: `dispenser` storage → `immutable` + 2-arg constructor + `changeDispenser` removed.**
- `addNominee` / `removeNominee` guard the dispenser call with `if (localDispenser != address(0))`, so a zero-dispenser (general-purpose) deployment is safe.
- The storage-layout change is a non-issue here — this is a fresh, non-upgradeable redeploy (no proxy), so there is no prior layout to stay compatible with.
- `scripts/deployment/deploy_23_vote_weighting.js` correctly reflects the ordering constraint (`deploy(veOLAS, dispenser)`, dispenser must be deployed beforehand or `AddressZero`).
- This is a reasonable hardening — it removes the owner-controlled dispenser-repoint vector.

**One deploy-time caveat (not a blocker):** because `dispenser` is now immutable and `changeDispenser` is gone, a deployment made with a wrong or zero dispenser address cannot be corrected in place — it requires a full redeploy. So the deployment config's `dispenserAddress` must be correct at construction; worth a checklist item in the deploy runbook.

Test files were updated for the new 2-arg constructor (`test/VoteWeighting.js`, forge tests, `MockDispenser`) — consistent with the change.

Verdict unchanged: **PASS.** The #8 fix stands and the immutable-dispenser refactor is sound.

DavidMinarsch:
## Review: `removeNominee` reconciliation

Verified locally: compiles under 0.8.30; **17/17 forge unit tests pass**; 3 of them genuinely fail on pre-fix logic (real regression coverage); Hardhat **175 passing** (body says 174).

**The core fix is sound.** Traced the accounting rather than taking the description at face value:

- `_getSum()` / `_getWeight()` return the *next-week* bias and have already consumed `changesSum[nextTime]` / `changesWeight[hash][nextTime]` and written `pointsSum[nextTime]` / `pointsWeight[hash][nextTime]` before `removeNominee` reads them. So `nomineeSlope` and `pointsSum[nextTime].slope` are on the same footing, and starting the strip loop at `nextTime + WEEK` is correct — not off-by-one.
- `changesSum[t] == Σₙ changesWeight[n][t]` is maintained exactly by `voteForNomineeWeights`, so subtracting the nominee's own entries is precise and leaves co-tenant nominees sharing that week untouched. The central claim holds.
- Horizon claim verified: `MAX_NUM_WEEKS` = 250 weeks vs veOLAS `MAXTIME` = 4×365d ≈ 208.6 weeks. Genuinely covers the maximum lock.
- `id != numNominees` (`VoteWeighting.sol:661`) is correct for last / first / single-element removal.
- Stripping the aggregate writes out of `revokeRemovedNomineeVotingPower` is right given removal now fully reconciles — no double-subtract.

---

### 1. [High — operational] Immutable `dispenser` is scope creep with a circular deploy and a silent footgun

Not one of #8 / #11 / #18 / #19 / #20. Consequences:

- **Circular deploy order.** `VoteWeighting(ve, dispenser)` needs the Dispenser first, but tokenomics `Dispenser.sol:346` rejects a zero `_voteWeighting`. A clean deploy therefore requires Dispenser(placeholder VW) → VoteWeighting → `Dispenser.changeManagers(0, 0, newVW)`. Neither deploy script documents this; the shell script only says "the Dispenser must be deployed BEFORE VoteWeighting".
- **Footgun.** `deploy_23_vote_weighting.sh` defaults a missing `dispenserAddress` to the zero address and prints it *in green* as though it succeeded. On mainnet that permanently bricks the Dispenser wiring with no recovery but redeployment. This should be a hard error unless an explicit no-dispenser flag is passed.
- Any future Dispenser redeploy now forces a VoteWeighting redeploy, and vote state cannot be migrated. Tokenomics #309 (Dispenser behind a proxy) is presumably the mitigation — worth stating explicitly here, since the two PRs are now load-bearing for each other.
- Credit where due: this closes recorded finding **#9** ("set up a dispenser contract immediately after deploying VoteWeighting and ensure that no nominees were added or removed before that") by construction. The PR doesn't claim it.

### 2. [Medium] No test exercises the actual revert-DoS

`test_RemoveWithoutUnvote_NoCheckpointDoS` establishes the *precondition* — on pre-fix logic it fails with sum `580e18` < surviving nominee weight `762e18`, so the next `removeNominee` would underflow — but it never executes the reverting call, and the 40-week `checkpoint()` loop does **not** revert on the pre-fix build.

The severe half of finding #8 — `_getSum()` itself reverting and bricking `nomineeRelativeWeightWrite` (and with it the Dispenser) — is not reproduced at all. That path requires the aggregate to clamp to zero, then fresh votes to lift `bias` above `dBias` while a phantom `changesSum[t]` still exceeds `pointsSum.slope`. Worth a `vm.expectRevert` case against the pre-fix build, or a note recording that it was reproduced out-of-band; otherwise there is no evidence the headline liveness bug is closed rather than only the drift.

Separately, `test_RemoveNominee_ExpiredVoter_NoUnderflow` **passes** on the pre-fix contract — despite the name it is not a regression test.

### 3. [Medium] The `1e18` clamp bounds each nominee, not the sum

Nothing enforces `Σ relativeWeight ≤ 1e18` across nominees. The Dispenser applies these weights per-nominee against a shared incentive pool (`Dispenser.sol:962`, `:1234`), so residual drift that inflates several nominees at once still over-allocates. #18 is *bounded*, not closed — worth saying so in the body.

### 4. [Medium] Drift now fails silently

The per-nominee walk clamps (`bias <= dBias → bias = 0, slope = 0`) independently of the aggregate walk, so `pointsSum.slope` can still diverge from `Σ pointsWeight.slope`. The new `_maxAndSub` guards convert what was a loud revert into quiet under-counting. That is the right trade, but there is no event or invariant check to detect it. The off-chain monitoring listed as optional in #8's "Workarounds" item 3 should become a launch requirement.

### 5. [Low — fix before merge] Stale ABI

`abis/0.8.25/VoteWeighting.json` still ships the 1-arg constructor plus `changeDispenser` / `DispenserUpdated`. The contract is now `^0.8.30` and `abis/0.8.30/` already exists. CI does not check this and downstream consumers (tokenomics deploy scripts, frontends) read these files.

### 6. [Low] The new forge tests never run in CI

`.github/workflows/workflow.yaml` runs eslint, solhint, `hardhat test` and `hardhat coverage` — zero `forge` references. The assurance case for this PR rests entirely on 19 forge tests that CI will never execute, so they will silently rot. Pre-existing gap, but this PR is what makes it matter. The fork tests can be gated on an RPC secret.

### 7. [Low] `docs/Vulnerabilities_list_governance.md` now contradicts the code

#20 still reads "Why this is not fixed: `VoteWeighting` is not upgradeable. Fixed on a future redeployment" and prescribes a *different* fix (add `_getSum()` / `_getWeight()` at the start of revoke) than the one shipped (remove the aggregate writes entirely — better, because it is idempotent). #8 is recorded as severity **Low** while this PR treats it as a checkpoint-bricking DoS. That file is the governance-facing artifact of record; leaving it stale through a security redeploy is exactly the drift that bites at audit time. It also references `retrieveRemovedNomineeVotingPower()`, which is not the contract's function name.

### 8. [Low] Deploy-script sanity checks don't fail

`deploy_23_vote_weighting.sh` echoes ve / dispenser / owner with "(must be …)" but never compares or exits non-zero on mismatch. For immutable slots, where a wrong argument means redeploying, make those hard assertions.

### 9. [Nit]

- The strip loop always runs 250 iterations even for a nominee with no scheduled changes (~525k gas in cold SLOADs alone). Tracking a per-nominee maximum scheduled `end` at vote time would allow an early break. Owner-only and rare, so optional — but the ~715k figure in the body is a floor, not a worst case.
- `verify_23_vote_weighting.js` has no trailing newline.

---

**Not a code issue, but the call that needs sign-off:** this is a fresh non-upgradeable redeploy with no described migration for existing nominees, votes, or removed-nominee history — voters must re-vote and the Dispenser must be re-pointed.

**Verdict:** the accounting redesign is correct and I'd approve it on merit. Blocking on **#1** (deploy-script footgun + documented ordering) and **#5**. **#2** should be a condition of the external re-audit already flagged in the body.

kupermind:
**Converged tokenomics-side view** (updating my earlier note now that #1 is resolved in code and #3/#4 are verified against the contract).

### #1 — resolved: no placeholder, no circular deploy

Implemented on the tokenomics side (Dispenser rework #314): `Dispenser.initialize` now accepts a **zero** `voteWeighting` (the Treasury zero-check stays). The deploy is therefore clean and non-circular:

1. deploy Dispenser impl + `DispenserProxy` (initialized with `voteWeighting = 0`, `StakingIncentivesPaused`);
2. deploy `VoteWeighting(ve, dispenserProxy)` — binding the now-known **proxy** address;
3. `changeManagers(0, realVW)` wires the real VoteWeighting while paused (the setter requires the paused state), then set deposit processors and unpause.

No mock / placeholder / old-VW address is used, and a fresh proxy can never be wired to an implementation address. A `changeManagers` wiring script was added (forge/cast; EOA-owned at deploy, a governance proposal once ownership moves to the Timelock) that hard-errors on a missing/zero VoteWeighting and asserts the result. This matches `deploy_23_vote_weighting.sh`'s documented "Dispenser before VoteWeighting" order — so the remaining ask stays on that script: its zero-default `dispenserAddress` (printed in green) should hard-error and assert the value is the **proxy** address.

On "any future Dispenser redeploy forces a VoteWeighting redeploy": with the Dispenser behind a stable proxy, future Dispenser fixes are `changeImplementation` swaps at the same address, so VoteWeighting's immutable never needs changing. #309 is the mitigation.

### #3 / #4 — verified: the over-mint is a false positive

Checked against your own `_nomineeRelativeWeight` — it's already handled inside VoteWeighting:

```solidity
weight = 1e18 * nomineeWeight / totalSum;
// Cap at 1.0 (1e18) ... defending the Dispenser consumer against any residual accounting drift
if (weight > 1e18) { weight = 1e18; }
```

- **A single nominee's weight is hard-capped at 1e18** — added explicitly (per the comment) to defend the Dispenser consumer. No individual weight can exceed 100%.
- **The sum can only exceed 100% by rounding dust.** `Σ relativeWeight = Σ pointsWeight[n].bias / pointsSum.bias`, and the accounting keeps `pointsSum.bias == Σ pointsWeight[n].bias` (lockstep voting, `changesSum == Σ changesWeight`, decay synchronized because a nominee's bias reaches zero at exactly the week its scheduled slope change fires). The residual clamp-asymmetry from #4 is dust-scale, not a doubling.
- **The Dispenser caps again per nominee** at `maxStakingIncentive`.

Per-nominee cap in VoteWeighting + the maintained sum invariant + the Dispenser's own per-nominee cap ⇒ no reachable path to a material over-mint. So I'd treat the over-allocation in #3 as a **false positive** and make **no Dispenser change**. The `pointsSum.slope` vs `Σ pointsWeight.slope` monitor from #4 is still reasonable hygiene, but it's belt-and-braces, not a tokenomics-side launch blocker.

(This supersedes my earlier note, which over-weighted the residual-drift risk before I'd confirmed the explicit 1e18 cap.)

The remaining items (#2 / #5 / #6 / #7 / #8 / #9) are governance-internal.

kupermind:
## Review follow-ups addressed (`70cf7ab`)

Thanks for the thorough pass. Actioned the code/repo items; the analysis-only items are answered inline.

**#1 — deploy footgun → hard failure (fixed).** `deploy_23_vote_weighting.sh` (and the JS twin) no longer default a missing/zero `dispenserAddress` to `address(0)`: it now prints red and `exit 1`. The contract still accepts a zero dispenser to run standalone, but that is not this deployment's case, so the scripts refuse it. Combined with **#8** below, the immutable-slot path is now fail-loud end to end. The circular-deploy half was already resolved on the tokenomics side (zero-`voteWeighting` init + Dispenser behind a stable proxy), so ordering is clean and non-circular.

**#8 — deploy-script sanity checks now assert (fixed).** The post-deploy `ve` / `dispenser` / `owner` echoes became hard assertions (case-insensitive compare, `exit 1` on mismatch). For immutable slots a wrong arg means a redeploy, so these fail the run rather than printing "(must be …)".

**#5 — stale ABI (fixed).** Added `abis/0.8.30/VoteWeighting.json` (2-arg ctor, no `changeDispenser`/`DispenserUpdated`) and removed the stale `abis/0.8.25/VoteWeighting.json`.

**#7 — vuln doc drift (fixed).** Reworded the "Why this is not fixed / future redeployment" stanzas for the VoteWeighting findings this redeploy closes (#8, #11, #18, #19, #20) to "addressed in the pending VoteWeighting redeployment" (implemented; takes effect on deploy + governance adoption). Fixed the `retrieveRemovedNomineeVotingPower` → `revokeRemovedNomineeVotingPower` typo. Left the GuardCM/verifier stanzas (#16, #17) untouched — not VW, not fixed here. The #8 "Low" severity label is the auditor's call; left as-is.

**#9 — nit (fixed).** Trailing newline added to `verify_23_vote_weighting.js`. The 250-iteration strip-loop early-break is left as-is deliberately: it would add a storage write to `voteForNomineeWeights` and logic to the just-audited `removeNominee` for an owner-only, rare gas saving — not worth re-opening the audited accounting.

**#3 — over-allocation: agreed false positive.** With the maintained invariant `pointsSum.bias == Σ pointsWeight[n].bias`, each `nomineeBias/totalSum` rounds down, so `Σ relativeWeight ≤ 1e18` — the sum drifts below 1.0 from dust, not above. The only path to `Σ > 1e18` is `pointsSum.bias` under-counting (the `_maxAndSub` saturation in #4), which only fires from an already-drifted state; on a reconciled build the subtraction is exact and nothing saturates. The per-nominee `1e18` cap plus the Dispenser's own per-nominee `maxStakingIncentive` cap close the residual. Recording as a false positive; no Dispenser change.

**#4 — silent drift: acknowledged, deferred (not a code defect).** The `_maxAndSub(a,b) = a>b?a-b:0` clamp trades a loud revert for a quiet under-count — the correct liveness-over-loudness call, but there's no on-chain signal. The minimal hardening is a diagnostic event emitted from `removeNominee` only when one of its `_maxAndSub` calls actually saturates (`oldWeight > oldSum`, `nomineeSlope > pointsSum.slope`, or `dSlope > changesSum[t]`); off-chain monitoring subscribes instead of reconstructing `Σ pointsWeight.slope`. Since it carries no state change and would re-open the audit surface, it's a follow-up rather than part of this PR. The `pointsSum.slope` vs `Σ pointsWeight.slope` monitor stays a launch requirement.

**#2 — test gap: agreed, tracked for the re-audit.** The existing test establishes the pre-fix precondition but never executes the reverting call, and the `_getSum()`-reverting half isn't reproduced. Adding a `vm.expectRevert` against the pre-fix build belongs with the external re-audit already flagged in the PR body.

**#6 — CI never runs forge: false positive for this PR.** Pre-existing gap (CI runs eslint/solhint/hardhat only, never forge) — not a regression introduced here.

77ph:
## Audit verdict — post-review-round (recorded in `audits/internal21/README.md`)

Independent adjudication of the review round + follow-up `70cf7ab`. Full write-up is in `audits/internal21` (updated on this branch).

**Delta verified first-hand.** `VoteWeighting.sol` is **byte-unchanged** since the accounting was verified — the net diff `8326d36…70cf7ab` does not contain the contract. The only deltas are the ABI regen (`0.8.25→0.8.30`), the vuln-doc reword, the deploy/verify scripts, and an unrelated `main` merge (proposal 12). The audited logic is intact; the follow-up hardens deploy + fixes ABI/docs only.

**Core fix.** Sound, now independently re-confirmed by two hand-traces converging on the same facts: `nextTime` footing (no off-by-one), `changesSum[t] == Σₙ changesWeight[n][t]` maintained exactly, strip horizon `MAX_NUM_WEEKS = 250 wk` > veOLAS `MAXTIME ≈ 208.6 wk`, and `id != numNominees` correct for last/first/single removal.

**On the review items**
- Resolved by `70cf7ab` (verified): #1 deploy footgun → hard-fail + circular-deploy resolved tokenomics-side (zero-`voteWeighting` init + proxy); #5 ABI; #8 deploy asserts; #7 doc reword; #9 newline.
- Concur: **#3 over-allocation is a false positive** (per-nominee `1e18` cap + `pointsSum.bias == Σ pointsWeight.bias` ⇒ `Σ relativeWeight ≤ 1e18` + Dispenser per-nominee cap); **#4** defer the diagnostic event, but keep the `pointsSum.slope` vs `Σ pointsWeight.slope` monitor as an **enforced** launch requirement.
- Would add / not lose:
  - **#2** — the headline `_getSum`-revert DoS is not reproduced by any test (the existing case only sets up the drift precondition). The flagged external re-audit should add a `vm.expectRevert` against the pre-fix build as a hard condition; and `test_RemoveNominee_ExpiredVoter_NoUnderflow` (passes on pre-fix) should be renamed — it is not a regression test.
  - **#6** — not a false positive but a correctly-scoped pre-existing gap this PR makes material: the assurance case rests on forge tests CI never runs. Wiring forge into CI (fork tests RPC-gated) is a launch condition.
  - **finding #8 severity** — a permanent checkpoint-DoS halting the weekly walk + the Dispenser reward path is at least **Medium**, not Low; worth correcting in the vulnerabilities list of record.
- **Coupling:** #2/#3/#4 reduce to one question — is the reconciliation exact? The re-audit closing #2 with a real revert-repro simultaneously firms up #3 and #4.

**Verdict: approve on code merit** — the accounting redesign is correct and byte-unchanged since verification; the follow-up resolved every blocking operational item without touching the audited logic. Ship gated on: (1) external re-audit that adds the #2 revert-repro; (2) the drift monitor as an enforced launch requirement; (3) forge → CI; (4) the finding-#8 severity correction; (5) the cross-repo lock — this PR and tokenomics #314/#309 are mutually load-bearing, neither ships without the other; (6) the test rename.

77ph (dismissed):

DavidMinarsch:
## Re-review after `70cf7ab` + internal21 update

Verified first-hand: `VoteWeighting.sol` is **byte-unchanged** since the accounting verification (`git diff 8326d36 70cf7ab -- contracts/VoteWeighting.sol` is empty — the audit's claim holds), and `forge test --match-contract VoteWeightingTest` is still **17/17**.

### Closed — verified

- **#1 deploy footgun** — both `deploy_23_vote_weighting.js` and `.sh` now hard-fail on a missing or zero `dispenserAddress` instead of silently defaulting. Resolved.
- **#5 stale ABI** — regenerated into `abis/0.8.30/`, `abis/0.8.25/` removed. I diffed the committed ABI against a fresh `forge build`: **65/65 entries identical**, 2-arg constructor, no `changeDispenser` / `DispenserUpdated`. Resolved.
- **#8 deploy sanity checks** — now hard assertions with case-insensitive comparison and `exit 1` on mismatch, rather than `echo`. Resolved.
- **#9** — trailing newline. Resolved.

### Withdrawn — #3 was wrong

The auditor's rebuttal is sound and I withdraw the item. Tracing it: the `Σ relativeWeight ≤ 1e18` bound does not come from the clamp at all — it comes from `pointsSum.bias == Σ pointsWeight[n].bias` together with the per-nominee round-down in `1e18 * nomineeWeight / totalSum`. Post-fix that invariant holds: at the week a nominee fully decays, its own walk takes the clamp branch and zeroes `slope`, while the aggregate consumes the matching `changesSum[t]` entry, so both sides drop by the same amount. The over-allocation direction is not reachable and no Dispenser-side change is needed.

My framing ("the clamp bounds each nominee but not the sum") was literally true about the clamp but pointed at a risk that does not exist. The audit's coupling note is also right that this safety rests on the reconciliation being exact — which is precisely what the missing #2 test would prove.

---

### New — [Medium] item #20 in the vulnerabilities doc now asserts something false about the shipped code

`70cf7ab` added to §20:

> **Status — addressed in the pending VoteWeighting redeployment.** … the **checkpoint-advance fix described below** is implemented and takes effect once the new VoteWeighting is deployed …

The **Fix on redeploy** text immediately below is unchanged and still reads:

> Call `_getSum()` and `_getWeight(account, chainId)` at the start of `revokeRemovedNomineeVotingPower`, mirroring the pattern in `voteForNomineeWeights`.

The shipped contract does neither. `revokeRemovedNomineeVotingPower` contains no `_getSum` / `_getWeight` call — only the comments explaining why the aggregate is deliberately left untouched. For comparison, §18's ("Clamp at the return site: `if (weight > 1e18) weight = 1e18;`") and §19's ("Change the guard to `if (id != numNominees)`") both match the shipped code exactly. §20's does not.

This is a regression introduced by the follow-up: before, the entry said "not fixed" — wrong, but harmless. It now claims a specific implementation that does not exist, in the governance file of record, heading into an external re-audit. Please reword to the design that actually shipped: revoke is reduced to pure per-user bookkeeping, the aggregate (bias, slope and future `changesSum` entries) is fully reconciled inside `removeNominee`, so there is no stale next-week checkpoint slot left to advance — which is why the fix is *by construction* rather than by mirroring `voteForNomineeWeights`.

### New — [Medium] internal21's adjudication of #1 rests on a mechanism that does not exist

The update records the circular deploy as:

> #1 deploy footgun → hard-fail scripts + the circular-deploy resolved tokenomics-side (**zero-`voteWeighting` init** + Dispenser behind a stable proxy, #314/#309)

and repeats it in ship gate (5) ("the zero-`voteWeighting` init + proxy remove the circular deploy").

There is no zero-`voteWeighting` init. `Dispenser.initialize` reverts `ZeroAddress()` when `_voteWeighting == address(0)` — `autonolas-tokenomics/contracts/Dispenser.sol:372-373` on the #315 head:

```solidity
// Check for at least one zero contract address
if (_treasury == address(0) || _voteWeighting == address(0)) {
    revert ZeroAddress();
}
```

The proxy does give a stable Dispenser *address*, which is the genuinely useful half — but it does not break the cycle, because VoteWeighting's constructor needs the Dispenser address and the Dispenser's initializer needs a non-zero VoteWeighting.

The cycle is real and has to be broken operationally:
1. deploy the Dispenser implementation + proxy, initializing with a **placeholder non-zero** `voteWeighting` (the currently-live VoteWeighting works);
2. deploy the new VoteWeighting with `_dispenser` = the Dispenser **proxy** address;
3. `Dispenser.changeManagers(address(0), newVoteWeighting)`.

So the deploy-footgun half of #1 is genuinely closed, but the **ordering** half is not — and it is now recorded as closed in the audit of record, which is a worse state than it being open. Worth correcting the internal21 text and writing the three steps into the deployment runbook, since gate (5) already (correctly) locks this PR and tokenomics #309/#314 together.

### New — [Low] `verify_23_vote_weighting.js` kept the zero fallback

Both deploy scripts now refuse a missing/zero dispenser, but the verification script still has:

```js
parsedData.dispenserAddress || "0x0000000000000000000000000000000000000000"
```

If the globals key is absent, Etherscan verification silently encodes the wrong constructor argument. Same footgun class as #1 — one file missed in the sweep.

---

### Still open, and we agree

#2 (no test exercises the actual revert-DoS), #6 (forge tests never run in CI), the #8 **Low** severity label, and the `test_RemoveNominee_ExpiredVoter_NoUnderflow` rename are all carried as ship gates in the internal21 verdict. Both the severity and the test name are still unchanged in the tree — tracked, not defects.

For the record: the adjudication is good faith. #2 and #6 were escalated rather than dismissed, and the severity dissent runs against the authors' own convenience. The two errors above are in the paperwork, not the accounting — which remains correct and byte-unchanged.

77ph:
## Final consolidated verdict

Re-review is correct on all three new items — verified first-hand, conceded, and the record is corrected in `audits/internal21` (correction dated 2026-08-13). There is no substantive dispute left: we agree the accounting is correct and byte-unchanged, and the two errors found were in our paperwork, not the code.

### Conceded (verified first-hand)
- **No zero-`voteWeighting` init exists — the circular deploy is real.** `Dispenser.initialize` reverts `ZeroAddress()` when `_voteWeighting == address(0)` (`Dispenser.sol:373-374`, dispenser-rework head). Our internal21 carried that claim over from a description without checking the tokenomics repo; it is **withdrawn**. The Dispenser proxy provides a stable *address* (the useful half) but does not break the cycle. The real path is operational and belongs in the runbook: (1) deploy Dispenser impl+proxy with a **placeholder non-zero** `voteWeighting`; (2) deploy `VoteWeighting(ve, dispenserProxy)`; (3) `changeManagers(0, newVoteWeighting)`.
- **#7 re-opened — §20 now mis-describes the shipped fix.** `70cf7ab`'s reword left §20's *Status* ("checkpoint-advance fix … implemented") contradicting its *Fix on redeploy* ("call `_getSum`/`_getWeight` at the start of revoke") — which the shipped code deliberately does **not** do (revoke is pure per-user bookkeeping; the aggregate is reconciled inside `removeNominee`, by construction). §20 should be reworded to the design that shipped, matching how §18/§19 already describe the code. My earlier "#7 resolved" confirmed only that the file changed, not that the reword was accurate — corrected.
- **[Low] `verify_23_vote_weighting.js:8`** kept the `|| "0x00…0"` fallback — same footgun class as #1, one file missed in the sweep.

### Settled (agreed by both reviews)
- **Accounting is correct and byte-unchanged** — `git diff … -- contracts/VoteWeighting.sol` empty, 17/17. Two independent hand-traces converge.
- **#3 withdrawn** — the `Σ relativeWeight ≤ 1e18` bound follows from `pointsSum.bias == Σ pointsWeight.bias` + per-nominee round-down, not the clamp; no Dispenser change.
- **#1 deploy footgun / #5 ABI / #8 sanity checks / #9** — resolved by `70cf7ab`.
- **#2, #6, finding-#8 severity (≥ Medium), and the `test_RemoveNominee_ExpiredVoter_NoUnderflow` rename** — carried as ship gates, agreed by both.

### Final residual checklist (none touch the audited accounting)
1. External re-audit of `removeNominee`/revoke that **adds the #2 revert-repro** test (also firms up #3/#4).
2. Off-chain `pointsSum.slope` vs `Σ pointsWeight.slope` drift monitor as an enforced launch requirement.
3. Forge tests wired into CI (fork tests RPC-gated).
4. Finding-#8 severity corrected in the vulnerabilities list (≥ Medium, not Low).
5. Cross-repo lock — this PR and tokenomics #309/#314 are mutually load-bearing; neither ships without the other, with the 3-step deploy ordering above in the runbook.
6. Doc/script corrections: §20 reword (#7), the deploy-ordering runbook (replacing the withdrawn zero-VW claim), and the `verify_23` zero fallback.
7. Rename the mislabeled regression test.

**Verdict: approve on code merit — settled.** The accounting redesign is correct and byte-unchanged; every remaining item is administrative (tests, CI, docs, severity, deploy runbook). With the checklist above closed, this completes the audit from our side.

77ph (dismissed):

kupermind:
## Re-review items addressed (`5c7d869`)

Both new items were correct — fixed. And the record correction is well taken.

**§20 doc regression (#7 re-open) — fixed.** Right that the prior reword left §20's *Status* claiming a "checkpoint-advance fix" that the shipped code deliberately does not implement, contradicting the untouched *Fix on redeploy*. Reworded both to the design that actually shipped: the aggregate (bias, slope, and future `changesSum` entries) is fully reconciled inside `removeNominee`, so `revokeRemovedNomineeVotingPower` is pure per-user bookkeeping and does not touch `pointsSum`/`pointsWeight`/`changesSum`/`changesWeight` — there is no stale next-week slot to advance, so the fix is *by construction* (mirroring `voteForNomineeWeights` would double-subtract the removed nominee's slope). Now consistent with how §18/§19 describe the shipped code.

**`verify_23_vote_weighting.js` zero fallback — fixed.** Dropped the silent `|| 0x00..0` and made it throw on a missing/zero `dispenserAddress`, mirroring both deploy scripts, so Etherscan verification cannot encode the wrong immutable constructor arg.

**Circular deploy — concur, my earlier note was wrong.** Confirmed first-hand against tokenomics: `Dispenser.initialize` reverts `ZeroAddress()` on a zero `_voteWeighting` (`Dispenser.sol:373-374`), so there is no zero-`voteWeighting` init and the cycle is real. My earlier follow-up repeated that false claim — withdrawing it here. The proxy gives a stable Dispenser *address* (the useful half) but does not break the cycle; the operational path is (1) deploy Dispenser impl+proxy with a placeholder non-zero `voteWeighting`, (2) deploy `VoteWeighting(ve, dispenserProxy)`, (3) `changeManagers(0, newVoteWeighting)`. That belongs in the deploy runbook, tracked under the cross-repo ship gate.

**Still-open ship gates unchanged and agreed:** #2 revert-repro test, #6 forge→CI, the finding-#8 severity (≥ Medium), the `test_RemoveNominee_ExpiredVoter_NoUnderflow` rename, the drift monitor, and the deploy runbook. Carried as launch conditions, not defects in this PR.

DavidMinarsch:
## Follow-up review — `5c7d869` + `71a9adc`

Both items from the last round are resolved. Verified first-hand: `VoteWeighting.sol` remains **byte-unchanged** (`git diff 257c1cc..HEAD -- contracts/VoteWeighting.sol` is empty).

- **§20 corrected.** Now reads *"Fix on redeploy (as shipped)"* and describes the actual design — `removeNominee` reconciles bias, still-active slope and future `changesSum` entries; `revokeRemovedNomineeVotingPower` is reduced to pure per-user bookkeeping and deliberately does not touch the aggregate; the stale-slot condition is removed by construction. Checked against the contract: accurate. The added note that mirroring `voteForNomineeWeights` "would double-subtract the removed nominee's slope" is the right thing to record — it explains why the original recommendation is now actively wrong, not merely superseded.
- **`verify_23_vote_weighting.js`** now hard-fails on a missing/zero dispenser instead of silently encoding the zero address. Matches both deploy scripts.
- **internal21 CORRECTION** — recording a dated correction that withdraws the zero-`voteWeighting`-init claim, rather than silently editing the verdict, is the right handling for an audit of record. Keeping ship-gate (5) while correcting only its stated reason is also right: the gate was correct, the mechanism was not.

### New, minor — the correction is already stale

The CORRECTION prescribes breaking the circular deploy with a **placeholder non-zero** `voteWeighting`:

> 1. deploy the Dispenser implementation + proxy, initialised with a **placeholder non-zero** `voteWeighting` (the currently-live VoteWeighting address works);

That was accurate against the tokenomics code at the time it was written, but `autonolas-tokenomics` #314 has since landed `cf881af`, which removes the zero check from `Dispenser.initialize` specifically to break this cycle. The shipped procedure is now:

1. `deploy_07a` (implementation) → `deploy_07b` (proxy, **`voteWeighting = 0`**);
2. deploy `VoteWeighting(ve, dispenserProxy)`;
3. `script_dispenser_change_managers.sh` (sets the real Vote Weighting while still paused).

So the two documents of record — this repo's internal21 and the tokenomics deploy scripts — currently describe different procedures. One line to reconcile, but worth doing before the external re-audit reads both.

### Still open, unchanged

#2 (no test exercises the actual revert-DoS), #6 (forge outside CI), finding #8's **Low** severity label, and the `test_RemoveNominee_ExpiredVoter_NoUnderflow` rename remain carried as ship gates. No change expected in this PR — noting only so they stay visible.

PR #218, opened 2026-08-19
docs(deployment): document the dispenserAddress prerequisite for VoteWeighting

kupermind:
## What

`VoteWeighting` binds the Dispenser as an **immutable** constructor argument, but nothing in this repo documented where that address comes from. `scripts/deployment/globals_mainnet.json` ships with `dispenserAddress` unset (`null`), so `./scripts/deployment/deploy_23_vote_weighting.sh mainnet` aborts on the guard with no pointer to the fix.

## Changes

- **`scripts/deployment/README.md`** — new "Deployment of VoteWeighting" section: get-into-deployable-state steps, required env keys, the deploy command, and the push-back steps. It states the prerequisite explicitly: `dispenserAddress` must be set from the [autonolas-tokenomics](https://github.com/valory-xyz/autonolas-tokenomics) `globals_<network>.json` key **`dispenserProxyAddress`** — the `DispenserProxy`, not the implementation — and re-checked whenever new Dispenser contracts are deployed there.
- **`deploy_23_vote_weighting.sh`** — the same rule as a header `PREREQUISITE` block; the missing/zero-dispenser error message now names the source key. Documents that the script overwrites the previous `voteWeightingAddress`.
- **`deploy_23_vote_weighting.js` / `verify_23_vote_weighting.js`** — matching comments so the Hardhat deploy path and the verification args carry the same source of truth.

## Notes

- A stale dispenser address is only fixable by redeploying VoteWeighting, which is why the refresh is documented as a per-run check rather than a one-off.
- The runbook states VoteWeighting is **Ethereum mainnet only** (it binds L1 veOLAS and the L1 Dispenser), against `globals_mainnet.json` / `chainId` 1 — not an L2.

## Scope

Docs and off-chain deployment scripts only. **No contract source is touched**, so no on-chain bytecode implications.

## Checks

`eslint` and `solhint` both clean (warnings only, all pre-existing). `bash -n` passes on the shell script.

### DIFF
--- a/scripts/deployment/README.md
+++ b/scripts/deployment/README.md
@@ -61,6 +61,62 @@ npx hardhat run scripts/deployment/deploy_17_governorTwo.js --network network_ty
 Then, after successful deployment of two supplemental contracts, the last script gives the proposal payload necessary to finalize the deployment:
 `npx hardhat run scripts/deployment/deploy_18_governor_to_governorTwo.js --network network_type`.
 
+## Deployment of VoteWeighting
+
+`VoteWeighting` is an Ethereum **mainnet** contract: it binds `veOLAS` (L1-only) and the L1
+`Dispenser` immutably in its constructor, so it is deployed against `globals_mainnet.json`
+(`chainId` 1) and not on any L2.
+
+### Get into a deployable state
+
+```bash
+git clone --recursive git@github.com:valory-xyz/autonolas-governance.git
+cd autonolas-governance
+yarn
+foundryup
+forge build
+```
+
+Export the original-Etherscan `ETHERSCAN_API_KEY` (used by `forge verify-contract`) and the
+`ALCHEMY_API_KEY_MAINNET` key the script appends to `networkURL`.
+
+### REQUIRED before deploying — refresh `dispenserAddress`
+
+`scripts/deployment/globals_mainnet.json` ships with `dispenserAddress` unset (`null`), and
+`deploy_23_vote_weighting.sh` aborts on a missing/zero value by design. Set it to the current
+live Dispenser address, taken from the [autonolas-tokenomics](https://github.com/valory-xyz/autonolas-tokenomics)
+repo's `scripts/deployment/globals_mainnet.json` key **`dispenserProxyAddress`** — the
+`DispenserProxy`, i.e. the address everything wires to, not the Dispenser implementation.
+
+This has to be re-checked on **every** run: whenever new Dispenser contracts are deployed in
+the tokenomics repo, the proxy address changes, and VoteWeighting stores the dispenser as an
+**immutable** — a stale or wrong address can only be corrected by redeploying VoteWeighting.
+
+```bash
+# in autonolas-governance, with <dispenserProxyAddress> copied from the tokenomics globals
+jq '.dispenserAddress = "<dispenserProxyAddress>"' scripts/deployment/globals_mainnet.json > tmp \
+  && mv tmp scripts/deployment/globals_mainnet.json
+```
+
+### Deploy
+
+```bash
+./scripts/deployment/deploy_23_vote_weighting.sh mainnet
+```
+
+The script deploys `contracts/VoteWeighting.sol`, writes `voteWeightingAddress` back into
+`globals_mainnet.json` (**overwriting** the previous VoteWeighting address), asserts the
+deployed `ve()` / `dispenser()` / `owner()` against the expected values, and verifies the
+contract on Etherscan (plus Blockscout when `blockscoutURL` is set).
+
+### Push back the changes
+
+```bash
+git checkout -b vw_deployment
+git commit -am "chore: deployment of VoteWeighting contract"
+git push origin vw_deployment
+```
+
 ## Deployment of the Veto stack (Task 1)
 
 A cancel-only Veto-Governor bound to a dedicated Veto Timelock. Both contracts are
--- a/scripts/deployment/README.md
+++ b/scripts/deployment/README.md
@@ -80,23 +80,43 @@ forge build
 Export the original-Etherscan `ETHERSCAN_API_KEY` (used by `forge verify-contract`) and the
 `ALCHEMY_API_KEY_MAINNET` key the script appends to `networkURL`.
 
-### REQUIRED before deploying — refresh `dispenserAddress`
+### REQUIRED before deploying — set `dispenserAddress`
 
-`scripts/deployment/globals_mainnet.json` ships with `dispenserAddress` unset (`null`), and
-`deploy_23_vote_weighting.sh` aborts on a missing/zero value by design. Set it to the current
-live Dispenser address, taken from the [autonolas-tokenomics](https://github.com/valory-xyz/autonolas-tokenomics)
-repo's `scripts/deployment/globals_mainnet.json` key **`dispenserProxyAddress`** — the
-`DispenserProxy`, i.e. the address everything wires to, not the Dispenser implementation.
+`scripts/deployment/globals_mainnet.json` has **no `dispenserAddress` key**, and
+`deploy_23_vote_weighting.sh` aborts on a missing/zero value by design. Add it, set to the
+**current live Dispenser** address from the
+[autonolas-tokenomics](https://github.com/valory-xyz/autonolas-tokenomics) repo's
+`scripts/deployment/globals_<network>.json`.
 
-This has to be re-checked on **every** run: whenever new Dispenser contracts are deployed in
-the tokenomics repo, the proxy address changes, and VoteWeighting stores the dispenser as an
-**immutable** — a stale or wrong address can only be corrected by redeploying VoteWeighting.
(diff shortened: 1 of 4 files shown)


DavidMinarsch (commented):
Verified against both repos' globals and against mainnet state. The gap this fixes is real — `VoteWeighting` binds the dispenser immutably, a wrong value is only fixable by redeploying, and nothing documented where the address comes from. **But the key name it points operators at does not exist today, and following the instruction literally in the wrong order produces exactly the permanent brick the PR exists to prevent.**

## The transitional hazard

The doc says (in four places) to copy the tokenomics globals key **`dispenserProxyAddress`** — "the `DispenserProxy`, not the implementation".

That key **does not exist** in `autonolas-tokenomics/scripts/deployment/globals_mainnet.json` on `main` today. The only dispenser key there is:

```
dispenserAddress = 0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0
```

…and that is the currently correct value — mainnet `VoteWeighting(0x95418b46…).dispenser()` returns exactly `0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0`. So **as of today the right key to copy is `dispenserAddress`**, and an operator following this README would look for `dispenserProxyAddress`, find `null`, and be stuck.

The instruction is right for the *post-migration* world. `deploy_07b_dispenser_proxy.sh` in tokenomics states its own contract in its header:

```
#   dispenserAddress      : Dispenser implementation (written by deploy_07a_*)
#   dispenserProxyAddress : deployed proxy address (the live Dispenser address to wire everywhere, …)
```

So after the Dispenser migration lands, `dispenserAddress` **changes meaning** — from "the live Dispenser" to "the implementation". Which makes the failure mode sharp in both directions:

- **Before the migration:** `dispenserProxyAddress` is absent; the correct value lives in `dispenserAddress`.
- **After the migration:** copying `dispenserAddress` binds VoteWeighting immutably to an *implementation* address — unrecoverable except by redeploying, i.e. precisely the accident this PR is written to prevent.

A doc that names only one of the two, with no mention that the key's meaning flips, is riskiest exactly during the window we are in now. Suggest stating both states explicitly and tying it to the migration, e.g.:

> Copy the **live Dispenser** address from the tokenomics `globals_<network>.json`. Once the proxied Dispenser is deployed (`deploy_07b_dispenser_proxy.sh`), that is the **`dispenserProxyAddress`** key — and from that point `dispenserAddress` in that repo means the *implementation*, which must **never** be copied here. Before that migration, the live Dispenser is `dispenserAddress`.

This also wants a pointer to the tokenomics migration runbook (`docs/dispenser_migration_runbook_public.md`, step 9 deploys VoteWeighting against the step-8 proxy), since the two documents describe the same handoff from opposite sides.

## Worth considering — a pre-flight guard, not just prose

The script asserts `ve()` / `dispenser()` / `owner()` **after** deployment (lines 127–129). By then the immutable is set, so the assert reports the mistake rather than preventing it. Given the failure is unrecoverable, a cheap pre-flight before `forge create` would be proportionate: require the address to have non-empty code, and ideally that it answers a Dispenser-proxy call (e.g. `voteWeighting()`), which an implementation-behind-a-proxy generally will not answer with sane state. That converts "brick discovered after the fact" into "refuse to deploy".

## Verified correct

- The script does what the README says: writes `voteWeightingAddress` back (line 123, overwriting — correctly documented as such), and asserts `ve()`/`dispenser()`/`owner()` post-deploy.
- The mainnet-only framing checks out: `globals_mainnet.json` has `chainId` 1, and live `VoteWeighting.ve()` = `0x7e01A500805f8A52Fad229b3015AD130A332B7b3` (L1 veOLAS), `owner()` = `0x3C1fF68f…95FE` (the DAO Timelock).
- The guard-message and comment additions in `.sh` / `.js` / `verify_*.js` are consistent with each other.
- No contract source touched, so the "no bytecode implications" claim holds.

## Minor

`globals_mainnet.json` in this repo does not "ship with `dispenserAddress` unset (`null`)" — the key is **absent entirely**. Behaviourally identical (`jq -r` yields `"null"`, JS yields `undefined`, both guards fire; and the README's `jq '.dispenserAddress = …'` creates it), so nothing breaks — but "absent" is the accurate word if the README is meant to be read literally.

kupermind:
Thanks — all four points verified independently and addressed in 2cc5176.

**Confirmed the transitional hazard.** On the tokenomics `main`, `globals_mainnet.json` has only `dispenserAddress = 0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0`, no `dispenserProxyAddress`; live `VoteWeighting(0x95418b46…).dispenser()` returns exactly that address, and `Dispenser(0x5650…).voteWeighting()` returns `0x95418b46…`. So today the correct key is `dispenserAddress`, and the previous wording would have sent an operator looking for a key that isn't there.

The docs now state **both** states and the meaning flip, as a table in the README plus prose in the `.sh` header, the deploy JS and the verify JS:

| State of the tokenomics repo | Live Dispenser = key | Do **not** copy |
|---|---|---|
| Before the migration (today) | `dispenserAddress` | — (`dispenserProxyAddress` does not exist yet) |
| After `deploy_07b_dispenser_proxy.sh` has run | `dispenserProxyAddress` | `dispenserAddress` — from then on the implementation |

Added the pointer to the tokenomics Dispenser migration runbook (step 9 → the step-8 proxy) so the two documents describe the same handoff from both sides.

**Pre-flight guard added**, before `forge create`:

1. the dispenser must have contract code;
2. it must read a **non-zero `owner()`**.

(2) is the discriminator you were after. It works because the new `Dispenser`'s constructor sets immutables only — `owner` is assigned in `initialize()`, which the `DispenserProxy` constructor delegatecalls — so a bare implementation reads `owner() == 0`, while both the pre-migration live Dispenser (`owner()` = `0x3C1fF68f…95FE`, the DAO Timelock) and a post-migration proxy read a real owner.

One deviation from the suggestion: `voteWeighting()` **cannot** be the check. A freshly deployed `DispenserProxy` is initialized with `voteWeighting == 0` on purpose (VoteWeighting does not exist yet — it binds the proxy, and is wired in afterwards via `changeManagers`), so asserting on it would reject the intended flow. It is printed for information, not asserted.

**Minor fixed:** "ships with `dispenserAddress` unset (`null`)" → "has no `dispenserAddress` key" — the key is indeed absent (`jq 'has("dispenserAddress")'` → `false`).

Guard behaviour spot-checked against mainnet: live Dispenser passes (`owner()` non-zero, `voteWeighting()` printed); a contract with no `owner()` and an EOA (`cast code` → `0x`) are both refused.

DavidMinarsch (commented):
Re-reviewed `2cc5176f`. Both items addressed, and the pre-flight guard is correct — I verified its discriminator at source rather than taking the reasoning on trust.

## Verified

- **The two-state table is accurate.** Pre-migration, the tokenomics globals' live Dispenser is `dispenserAddress = 0x5650300fCBab43A0D7D02F8Cb5d0f039402593f0`, which matches mainnet `VoteWeighting(0x95418b46…).dispenser()` exactly. Post-migration, `deploy_07b_dispenser_proxy.sh`'s own header defines `dispenserAddress` as the implementation and `dispenserProxyAddress` as the proxy. The table matches both states, and the cross-link to the tokenomics runbook closes the loop from the other side.
- **The guard's discriminator is sound.** `Dispenser.sol`'s constructor sets implementation-bytecode immutables only (its own comment at `:333-334`), and `owner = msg.sender` lives in `initialize()` at `:390`, which the `DispenserProxy` constructor delegatecalls. So a bare implementation genuinely reads `owner() == 0`, while a proxy does not — the check separates exactly the two things that need separating.
- **No false positive today**: the live Dispenser has 18,684 bytes of code and `owner()` = `0x3C1fF68f…95FE` (the DAO Timelock), so the guard passes on the address an operator would use right now.
- **The `voteWeighting()` non-assertion is right.** A freshly deployed proxy is initialized with `voteWeighting == 0` by design — tokenomics runbook step 8 initializes with exactly that — so it cannot discriminate. Printing it for information is the right compromise.
- "has no `dispenserAddress` key" is now accurate.

## One residual, worth stating so nobody "fixes" it later

The pre-flight separates *implementation* from *live*, but it cannot separate *stale-live* from *current-live* — an old Dispenser and a new one both have code and a non-zero owner. Staleness is the other failure mode the doc names, and it stays uncaught.

The obvious cross-check would be to compare `dispenserAddress` against the on-chain `Tokenomics.dispenser()` / `Treasury.dispenser()`. **That would be wrong here**, and it is worth writing down why: in the migration sequence, VoteWeighting is deployed at step 9, while Tokenomics and Treasury are re-pointed at the new Dispenser only at steps 17–18. A wiring cross-check would therefore fail on every legitimate migration run, exactly when it is most needed.

So the guard stopping where it does is a deliberate boundary, not an omission — and the "re-check the value on every run" instruction is what carries the staleness case. A one-line note to that effect in the script header would stop a future contributor from adding the tempting-but-breaking check.

Nothing blocking.

PR #219, opened 2026-08-21
docs: unbounded Governor settings & unchecked Timelock replacement (Vulnerabilities list #25)

kupermind:
Adds a consolidated entry #25 to `Vulnerabilities_list_governance.md` covering four related self-brick configuration issues:

- `setProposalThreshold` unbounded → `propose()` permanently blocked
- `setVotingDelay` / `setVotingPeriod` unbounded → voting schedule pushed out indefinitely
- `updateGovernorDelay` unbounded → queued-proposal ETA unreachable, execution frozen
- `updateTimelock` without role-wiring validation → Governor orphaned from a Timelock it cannot drive

All four setters are `onlyGovernance` (caller must be the executor/Timelock), so none is externally reachable and none extracts value — the realistic trigger is an unsafe parameter in a legitimate proposal, not a permissionless attack. The entry records operator guidance (keep parameters bounded, wire Timelock roles before replacement, keep a recovery path available) and the redeploy fix (bound the settable ranges; validate role wiring in `updateTimelock`).

Docs-only; no contract source changed.

### DIFF
--- a/docs/Vulnerabilities_list_governance.md
+++ b/docs/Vulnerabilities_list_governance.md
@@ -28,6 +28,7 @@
 | 22 | [WormholeMessenger single sourceGovernor authentication](#22-wormholemessenger-single-sourcegovernor-authentication) | Informative |
 | 23 | [Timelock deployer-EOA admin bootstrap window](#23-timelock-deployer-eoa-admin-bootstrap-window) | Informative |
 | 24 | [FxERC20RootTunnel setFxChildTunnel is a permissionless one-shot](#24-fxerc20roottunnel-setfxchildtunnel-is-a-permissionless-one-shot) | Informative |
+| 25 | [Unbounded Governor settings and unchecked Timelock replacement can permanently brick governance](#25-unbounded-governor-settings-and-unchecked-timelock-replacement-can-permanently-brick-governance) | Medium |
 
 ## Involved contracts and level of the bugs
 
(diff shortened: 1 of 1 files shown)


DavidMinarsch (commented):
Checked the entry against the contracts it describes. The consolidation is sound and the `onlyGovernance` reasoning is correct — I verified all four setters — but two of the code claims are imprecise in ways that would mislead whoever implements the redeploy fix.

## Verified

`updateGovernorDelay` (`contracts/utils/GovernorTimelockControl.sol:189`) and `updateTimelock` (`:173`) are both `external virtual onlyGovernance`, and `setProposalThreshold` / `setVotingDelay` / `setVotingPeriod` come from OZ `GovernorSettings`, which gates all three the same way. So "none is externally reachable, the realistic trigger is operator error in a legitimate proposal" holds for all four. The ToC anchor resolves.

## `updateGovernorDelay` is not unbounded — it is bounded below

`_updateGovernorDelay` already validates:

```solidity
uint256 minDelay = _timelock.getMinDelay();
if (newGovernorDelay < minDelay) {
    revert Underflow(newGovernorDelay, minDelay);
}
```

The brick scenario the entry describes needs the *upper* direction, so the substance is right — there is no maximum. But "unbounded" as written is wrong in the lower direction, and it matters for the fix: "reject a `governorDelay` outside a configured min–max" reads as though both sides are missing, when the min side exists and is *dynamic* (it tracks `timelock.getMinDelay()`, which is itself separately updatable — the coupling item 13 is about). An implementer adding a static min–max could easily weaken that existing relative check. Suggest saying the floor exists and only a ceiling is needed.

## `updateTimelock` is worse than described — there is no validation at all

The entry says the replacement "is not validated against the required role wiring". It is not validated against *anything*:

```solidity
function _updateTimelock(TimelockController newTimelock) private {
    emit TimelockChange(address(_timelock), address(newTimelock));
    _timelock = newTimelock;
}
```

No role check, and **no zero-address check**. `updateTimelock(TimelockController(address(0)))` is accepted, which is a worse end state than "a Timelock it cannot drive" — every subsequent queue/execute path calls into a codeless address.

That is worth separating in the fix section, because the two mitigations are very different in size: a zero-address check is one line and closes the most catastrophic variant, while validating `PROPOSER` / `EXECUTOR` / `CANCELLER` wiring is a larger change. Recommending both, in that order, gives a cheap partial fix that does not depend on the full one landing.

## Severity — Medium is out of step with the document's own calibration

The ToC currently runs 15 Informative / 6 Low / 4 Medium, and the closest neighbours are all Informative:

- **#13** — `governorDelay` / timelock `minDelay` desynchronisation, the *same parameter* this entry says it compounds — **Informative**.
- **#24** — `setFxChildTunnel` is a **permissionless** one-shot — **Informative**.
- **#23** — Timelock deployer-EOA admin bootstrap window — **Informative**.

So the document rates a permissionless one-shot Informative, and rates #25 — which its own text argues is not externally reachable, has no value extraction, and is reachable only by operator error — a Medium. There is a defensible reason for the gap (these end states are *unrecoverable*, whereas #13's desync is correctable), but that rationale is not stated, and a reader comparing #13 with #25 will just see an inconsistency. Either state the unrecoverability basis explicitly in the entry, or align the severity with the neighbours.

## Minor

- The **Location** line omits `setVotingPeriod`, though the body covers it (`**setVotingDelay** (and setVotingPeriod)`). Since Location is what a reader greps, it should list all five functions.
- The **"Involved contracts and level of the bugs"** section still says the document "aims to point out some vulnerabilities in the contracts veOLAS, buOLAS, and VoteWeighting". That was already stale for #22–#24 (WormholeMessenger, Timelock, FxERC20RootTunnel) and #25 adds `GovernorOLAS` / `GovernorTimelockControl` to the drift. Not this PR's doing, but this PR is the fourth entry to fall outside that sentence — worth a one-line refresh while touching the file.

PR #225, opened 2026-08-31
proposal 13: cross-chain housekeeping (Mode bridge route, Mode staking allowlist, PolySafe retirement)

DavidMinarsch:
Three unrelated maintenance actions batched into one vote. All submitted as L1 Timelock calls; two carry a bridged payload.

**proposalId** `100619077063411664334557367612251066850008502834926962826958681194909586886778`
**descriptionHash** `0xfc97ad5c460cb572b31ee9f3c4bfb17e114055fdabfad7e56e0c8c07d30cdeb9`

| # | Chain | Route | Call |
|---|---|---|---|
| 0 | Ethereum | direct | `GuardCM.setBridgeMediatorL1BridgeParams` — register the Mode (34443) route |
| 1 | Mode | Mode L1CDM → mediator | `StakingVerifier.setImplementationsStatuses([V1], [false], true)` |
| 2 | Polygon | FxRoot → FxGovernorTunnel | `ServiceRegistryL2.changeMultisigPermission(PolySafeCreator, false)` |

## [0] Mode bridge route

Proposal 11 configured six chains but not Mode, so GuardCM holds `verifierL2 == address(0)` for the Mode L1 messenger and **every community-multisig transaction to Mode fails closed** — the only supported L2 where the CM cannot act. Mode is OP-stack and reuses the existing `ProcessBridgedDataOptimism` verifier. Adds a capability, removes none.

## [1] Mode staking allowlist

Drops a superseded V1 staking implementation. Nothing is unwound: Mode's `StakingFactory` has emitted **7** `InstanceCreated` events, all against a different implementation, **0** against this one.

> ⚠️ **`setCheck` must be `true`.** The contract assigns `implementationsCheck = setCheck` **unconditionally** — `false` would switch the allowlist off entirely and make `verifyImplementation()` return `true` for everything, the exact opposite of the intent. The Mode fork test asserts `implementationsCheck()` is still `true` afterwards; that is the load-bearing assertion.

## [2] PolySafe retirement — a product decision, not only a fix

Closes registries **#30** outright with no code change. But **PolySafe is not dormant**: 259 `MultisigCreated` events, most recent **2026-08-15**, though volume fell from 92 (April) to 3 (August). After execution, a service that would have deployed a PolySafe deploys an ordinary Safe. `GnosisSafeMultisig` stays whitelisted as the fallback and the Polygon fork test asserts it survives untouched. Reversible by re-whitelisting; in-flight PolySafe deploys revert meanwhile.

## ⚠️ Address collisions — the main review hazard

`0x9338b515…` is Polygon's `FxGovernorTunnel`, Mode's mediator **and** Gnosis's `ServiceRegistryL2`. `0x87c511c8…` is **both** Optimism's L2 messenger and Mode's `StakingVerifier`. A constant pointed at the wrong chain looks correct on review, so each was verified by calling a **chain-specific getter on the chain it is used on**:

| check | result |
|---|---|
| Mode `0x9338b515…`.`CDMContractProxyHome()` | `0x4200…0007` |
| Mode `0x9338b515…`.`sourceGovernor()` | L1 Timelock |
| Polygon `0x9338b515…`.`fxChild()` / `rootGovernor()` | `0x8397259c…` / L1 Timelock |
| Mode `0x87c511c8…`.`owner()` | the Mode mediator |

The annotator resolves chain from the **L1 entrypoint**, never the L2 address — so the same address correctly links to polygonscan in entry 2 and Mode's explorer in entry 1.

## Testing — 8 tests, all passing

**L1** (4): full propose → vote → queue → execute through the live GovernorOLAS; both bridged entries observed dispatching (`SentMessage` to the Mode mediator; `StateSynced` carrying `rootSender == Timelock`, `receiver == FxGovernorTunnel`); preconditions with an Optimism-route positive control so the Mode zero is a real zero; and a pin on descriptionHash + proposalId so a drifted description can't be submitted by accident. `execute()` = **2,548,065 gas** vs the EIP-7825 cap of 16,777,216.

**L2**: Polygon validated by **Tenderly** per convention — state diff shows `mapMultisigs[0xa749f605…]` **`true` → `false`**, with `MessageReceived(rootMessageSender = Timelock)`.

**Tenderly does not support Mode** — 34443 is absent from its public-networks list and `/simulate` returns HTTP 500. Rather than validate one leg more weakly than the other, both legs also get destination-chain fork tests replaying the proposal's own packed buffers, each with a negative test for a foreign message origin.

```bash
ETH_RPC=<rpc>     forge test --match-contract Proposal13HousekeepingTest -vvv
MODE_RPC=https://mainnet.mode.network forge test --match-contract Proposal13ModeLegTest -vvv
POLYGON_RPC=<rpc> forge test --match-contract Proposal13PolygonLegTest -vvv
```


kupermind:
## Review — proposal 13

Reviewed the calldata byte-by-byte, re-derived the pinned identifiers, re-verified every on-chain
claim against mainnet / Mode / Polygon, and re-ran all 8 tests. **The three entries encode exactly
what the description says**, and the artifacts are internally consistent. Three things worth fixing
before submission, plus some nits.

### Verified independently

| Check | Result |
|---|---|
| Selectors `0x1602c55c` / `0x3dbb202b` / `0xd3042d2b` / `0x0bea55ed` / `0xb4720477` / `0x82694b1d` | all match their claimed signatures |
| `keccak(description.txt)` (778 bytes, no trailing newline) | `0xfc97ad5c…0cdeb9` — matches the pinned descriptionHash |
| `keccak(abi.encode(targets, values, calldatas, descHash))` recomputed from the committed `calldata.json` | `0xde746123…33107a` = `1006190770…86886778` — matches the pinned proposalId |
| `node annotate.js` re-run against the committed inputs | regenerates `proposal_13.html` **byte-identically** |
| `setCheck` in entry [1]'s innermost payload | raw word is `0x…01` — `true`, as required |
| Packed buffers | `20+12+4+228 = 264 = 0x108` and `20+12+4+68 = 104 = 0x68` — both self-consistent |
| `GuardCM.owner()` | `0x3C1fF68f…95FE` (Timelock) |
| `GuardCM.mapBridgeMediatorL1BridgeParams(ModeL1CDM)` | `(0, 0, 0)`; Optimism control returns `(0xdCAFcCcC…, 0x87c511c8…, 10)` |
| Mode `L2CrossDomainMessenger(0x4200…0007).OTHER_MESSENGER()` | `0x95bDCA6c…` — **definitive** proof the L1 entrypoint is Mode's, from the Mode side |
| Mode `0x9338b515…` `CDMContractProxyHome()` / `sourceGovernor()` | `0x4200…0007` / the L1 Timelock |
| Mode `0x87c511c8…` `owner()` / `implementationsCheck()` / `mapImplementations(V1)` | mediator / `true` / `true` |
| Polygon `0x9338b515…` `fxChild()` / `rootGovernor()` | `0x8397259c…` / the L1 Timelock |
| Polygon `ServiceRegistryL2.mapMultisigs` for the PolySafe creator and `GnosisSafeMultisig` | `true` / `true` |
| **"0 of 7 proxies against V1"** | reproduced from Mode logs: exactly 7 `InstanceCreated`, all from StakingFactory `0x75D529FA…`, **all** with implementation `0xE49CB081…`, **none** against `0x88DE7346…` |
| Tests | 4/4 L1, 2/2 Mode, 2/2 Polygon pass; negative tests revert with `SourceGovernorOnly` / `RootGovernorOnly` |

One claim I could **not** re-verify: the 259 `MultisigCreated` / "most recent 2026-08-15" volume figures —
public Polygon RPCs cap `eth_getLogs` at 10k-block ranges and a full-history scan timed out. Not
load-bearing for safety, but it is un-reverified here.

---

### 1. `proposal_13.html` mislabels the Mode `StakingVerifier` — should fix

The generated HTML renders `0x87c511c8aE3fAF0063b3F3CF9C6ab96c4AA5C60c` as:

```
0x87c511c8aE3fAF0063b3F3CF9C6ab96c4AA5C60c  // OptimismMessenger (Optimism L2) · Mode
```

The chain suffix is right, the contract name is Optimism's messenger. `ADDR` in `annotate.js` is a
flat `address → name` map with no chain dimension, so the exact collision the PR calls out as "the
main review hazard" is present, unmitigated, in the one artifact voters actually read. Suggest
keying `ADDR` by `"<chainId>:<address>"` (or adding a per-chain override consulted first), so
`34443:0x87c511c8…` resolves to `StakingVerifier (Mode)`.

### 2. The annotator resolves the OP-stack chain from the L2 address, not the L1 entrypoint — should fix

The description says "The annotator resolves chain from the **L1 entrypoint**, never the L2 address".
For the FxPortal and AMB branches that is true (`137` and `100` are hardcoded), but the OP-stack
branch does the opposite — `annotate.js:264`:

```js
const [target, message, minGas] = abi.decode(["address", "bytes", "uint32"], args);
const cid = MSGR2CHAIN[lc(target)] || 1;   // target is the L2 receiver
```

It produces the right answer here only because `MSGR2CHAIN` happens to map `0x9338b515…` to `34443`
— i.e. the *same* key that is also Polygon's `FxGovernorTunnel`. The L1 entrypoint (`e.target`) is
already in scope in `decodeEntry`; deriving `cid` from an L1CDM → chainId map instead would make the
code match the stated (and correct) mitigation. Either fix the code or soften the claim.

### 3. Audit script drift, and a comment that contradicts this proposal — should fix

`scripts/audit_chains/audit_contracts_setup.js:431`:

```js
// Note: Mode (chain Id 34443) is intentionally not configured on this guard (being deprecated).
```

and the `bridgeParams` array immediately below omits Mode. After this proposal executes, that
comment is false and the setup audit silently skips the newly registered route — the one route with
no coverage anywhere. `globals_mainnet.json` already carries `modeL1CrossDomainMessengerAddress` and
`modeMessengerL2Address`, so adding the Mode row is a one-liner. It would also be worth addressing
in the PR description why "being deprecated" is superseded, since a reviewer who greps for Mode
finds that note first and it argues the opposite of entry [0].

### 4. Entry [0]'s actual capability is never asserted — nice to have

The tests assert the three storage words land. They never assert the thing entry [0] exists for: that
a Mode-bound CM transaction now passes `GuardCM.checkTransaction`.

Worth noting because I checked and it *would* pass — the CM allowlist already carries the Mode pairs
on the live guard:

| key | live value |
|---|---|
| (`0x3C1fF68f…` = `ServiceRegistryL2` on Mode, `drain()`, `34443`) | `true` |
| (`0x34C895f3…` = `ServiceRegistryTokenUtility` on Mode, `drain(address)`, `34443`) | `true` |

(Both confirmed on Mode: `serviceRegistry()` / `owner()` / `drainer()` resolve as expected.) So the
bridge route genuinely is the single missing link, and a test that pranks the CM through
`checkTransaction` with a `schedule()` to the Mode L1CDM — reverting before, passing after — would
upgrade the claim from "storage written" to "capability restored", and would catch a future
regression in either half. Adding the two allowlist entries to `test_preconditions` is the cheap
version.

Related, worth one line in the README: entry [0] also changes how the guard treats a **direct** L1 CM
call to the Mode L1CDM — previously verified against `block.chainid` and rejected, now routed through
`ProcessBridgedDataOptimism`. No capability is removed either way, but it is a second behavioural
delta from the same storage write.

### 5. Nothing pins `description.txt` / `calldata.json` to the builder — nit

`test_proposalIdMatchesCommittedDescription` pins the hash of the builder's `DESCRIPTION` constant,
which is exactly the drift the header warns about ("`description.txt` MUST match … byte-for-byte")
except that it does not check the file. Same for `calldata.json`, which is what the HTML is generated
from. Two `vm.readFile` assertions close both:

```solidity
assertEq(keccak256(bytes(vm.readFile("scripts/proposals/proposal_13/description.txt"))), dh);
// and per-entry: vm.parseJsonBytes(json, ".[0].calldata") == calldatas[0], etc.
```

Both files do match today (checked by hand) — this only makes it stay true.

### 6. Negative tests accept any revert — nit

`test_mode_rejectsForeignRootSender` / `test_polygon_rejectsForeignRootSender` use a bare
`vm.expectRevert()`, which would also pass if the call failed for an unrelated reason (a malformed
buffer, a mock that silently did not apply). They currently revert with
`SourceGovernorOnly(0x…BEEF, Timelock)` and `RootGovernorOnly(0x…BEEF, Timelock)` — pin those.

### 7. The Polygon test's default RPC is dead — nit

`vm.envOr("POLYGON_RPC", "https://polygon-rpc.com")` returns HTTP 401 today
(`API key disabled, reason: tenant disabled`), so both Polygon tests fail outright unless the env var
is set, while the Mode and L1 defaults work. `https://polygon-bor-rpc.publicnode.com` works.

### 8. Leftover categories in `annotate.js` — nit

`GROUP_ORDER` carries `dewhitelist` / `nominee` / a second `guard` heading from earlier proposals,
and every selector in `DEWHITELIST_SELS` is already matched by an earlier branch in `category()`, so
those arms are unreachable. Harmless, but it invites a wrong grouping the next time the file is
copied forward.

DavidMinarsch:
Thanks — all eight actioned, and two of them found real defects rather than nits. **`proposalId` and `descriptionHash` are unchanged; no calldata was touched.** 10/10 tests pass.

## 1 & 2 — the annotator (both were right, and worse together)

You caught the collision **inside the mitigation for the collision**. `ADDR` is a flat address→name map with no chain dimension, so the HTML labelled Mode's `StakingVerifier` as Optimism's messenger. Added `ADDR_BY_CHAIN`, keyed `chainId:address` and consulted first:

```
0x87c511c8…  // StakingVerifier (Mode) · Mode
0x9338b515…  // FxGovernorTunnel (Polygon) · Polygon      ← entry 2
0x9338b515…  // OptimismMessenger mediator (Mode) · Mode  ← entry 1
```

On (2) — you're right that the description claimed something the code didn't do. **I fixed the code rather than softening the claim.** The OP-stack branch now resolves the chain from `e.target` via `L1CDM2CHAIN` (the L1 messenger is unique per chain; the L2 receiver is not) and **throws** on an unknown messenger instead of silently defaulting to `1`. The old path gave the right answer here only by luck of which entry that ambiguous key happened to hold. `MSGR2CHAIN` is now unused and removed.

## 3 — audit script

Added the Mode row and replaced the comment. Worth being explicit about why "being deprecated" is superseded: leaving the route unset is *precisely* what made Mode the one supported L2 the CM could not reach, because the guard fails closed on an unset route. Also noted in the README so a reviewer who greps for Mode finds the resolution next to the note.

## 4 — capability assertion

Added, and it was worth doing. `test_L1_modeCmTransaction_rejectedBefore_acceptedAfter` drives a Mode-bound CM `schedule()` through `GuardCM.checkTransaction` — **rejected before, accepted after** — and asserts the allowlist half separately so a failure is attributable to the bridge route rather than a missing pair. Your allowlist reads reproduce.

One thing it caught immediately: my first draft **hardcoded the wrong CM address**. The live `multisig()` is `0x04c06323…`. It now reads the CM from the guard at runtime, since a stale constant there would have made the test assert against the wrong caller and pass for the wrong reason.

Split into helpers to stay under the stack ceiling rather than enabling `via_ir`.

Also added the direct-L1-call delta to the README, as you suggested.

## 5–8 — nits

- **5:** `test_committedArtifactsMatchTheBuilder` reads both files off disk. You're right that the old test pinned the builder against itself. Needed a read-scoped `fs_permissions` entry.
- **6:** both negative tests now pin `SourceGovernorOnly` / `RootGovernorOnly`.
- **7:** default is now `https://polygon-bor-rpc.publicnode.com`.
- **8:** dead categories and `DEWHITELIST_SELS` removed.

## On the one claim you couldn't re-verify

The 259 `MultisigCreated` / 2026-08-15 figures — public Polygon RPCs do cap `eth_getLogs`, so that's a fair flag. It isn't safety-load-bearing, as you say, but it *is* the evidence for calling PolySafe "not dormant", which is what makes entry [2] a product decision rather than routine cleanup. Happy to re-run it against an archive endpoint and post the block range if you'd like it independently checkable.

kupermind:
## Re-review — 3812471

All eight are addressed, and I re-verified each fix rather than taking the summary at face value.
**The proposal itself is untouched**, so the pinned identifiers cannot have moved:

```
$ git diff --name-only 10bac7f 3812471
foundry.toml
scripts/audit_chains/audit_contracts_setup.js
scripts/proposals/proposal_13/{README.md, annotate.js, proposal_13.html}
test/proposals/Proposal13{Housekeeping,L2Legs}.t.sol
```

`calldata.json`, `description.txt` and `Proposal13Housekeeping.s.sol` are byte-identical across the two
commits — the fixes are confined to tooling, tests and docs. 10/10 tests pass here.

### Fixes verified

| # | Fix | How I checked it |
|---|---|---|
| 1 | `ADDR_BY_CHAIN`, consulted before the flat map | In the regenerated HTML the same 20 bytes now render three different ways: `0x87c511c8…` → `StakingVerifier (Mode)`, `0x9338b515…` → `OptimismMessenger mediator (Mode)` in entry 1 and `FxGovernorTunnel (Polygon)` in entry 2 |
| 2 | OP-stack chain from `e.target` via `L1CDM2CHAIN`, throws on unknown | All four entries match `globals_mainnet.json` exactly (`0x25ace71c…`→10, `0x866E82a6…`→8453, `0x1AC1181f…`→42220, `0x95bDCA6c…`→34443); `MSGR2CHAIN` is gone (0 references) |
| 3 | Mode row added to `bridgeParams`, comment replaced | Row reads the two `globals_mainnet.json` keys that already exist |
| 4 | Capability test | Traced it — the "after" leg genuinely reaches `0xdCAFcCcC…::processBridgeData(…, 0x9338b515…, 34443) [delegatecall]`. It is a real end-to-end assertion, not a storage re-read |
| 5 | `test_committedArtifactsMatchTheBuilder` | Reads both files off disk; `fs_permissions` is read-scoped to `./scripts/proposals` |
| 6 | Pinned `SourceGovernorOnly` / `RootGovernorOnly` | Both pass with the pinned data |
| 7 | Polygon default RPC | `polygon-bor-rpc.publicnode.com` works without env vars |
| 8 | Dead categories removed | `DEWHITELIST_SELS` at 0 references; `category()` now maps exactly the three selectors present |
| — | HTML reproducibility | `node annotate.js "Proposal 13 — cross-chain housekeeping"` (the command in README:130) regenerates `proposal_13.html` **byte-identically** |
| — | CI-runnable checks | `eslint` 0 errors, `solhint` 0 errors |

Reading the CM off the guard instead of hardcoding it was the right call, and finding the stale constant
that way is a good argument for the test existing at all.

---

### A. The capability test's "before" leg is a bare `vm.expectRevert()` — should fix

This is finding 6 reappearing in the one test where it matters most. `Proposal13Housekeeping.t.sol:227`:

```solidity
vm.prank(cm);
vm.expectRevert();          // accepts any revert
_checkCmTransaction(scheduleCall, cm);
```

The whole weight of "rejected before, accepted after" rests on that leg failing *for the stated reason*.
As written it would pass just as happily if the schedule call were malformed, the ABI drifted, or `cm`
resolved to something unexpected — in which case the "before" half proves nothing and the test still goes
green. The L2 tests were tightened for exactly this reason in the same commit.

It currently reverts with:

```
NotAuthorized(0x95bDCA6c8EdEB69C98Bd5bd17660BaCef1298A6f, 0x3dbb202b, 1)
```

Worth pinning, and the value is more than pedantic: that revert is *itself* the evidence for the
direct-L1 delta now documented in the README — chainId `1`, selector `sendMessage`, i.e. the guard
falling through to the L1 path because no route exists. Pinning it documents the pre-state precisely:

```solidity
vm.expectRevert(
    abi.encodeWithSignature("NotAuthorized(address,bytes4,uint256)", MODE_L1CDM, bytes4(0x3dbb202b), uint256(1))
);
```

Signature checked against the declaration — `VerifyData.sol:8`:
`error NotAuthorized(address target, bytes4 selector, uint256 chainId);`

### B. `category()` still derives the chain from the selector — should fix

Finding 2 was fixed in `addrSpan`, but the grouping logic one function above it still does the old thing —
`annotate.js:213`:

```js
function category(sel) {
    if (sel === "0x3dbb202b") return "mode";
```

`0x3dbb202b` is `sendMessage` on *any* OP-stack L1 messenger, not Mode's. The heading it selects is
literally chain-named ("Mode: de-allowlist…"), so the moment a proposal carries an Optimism, Base or Celo
leg alongside a Mode one, both land under the Mode heading — or worse, a lone Base entry is titled Mode.
This is the same class of defect and the same blast radius as finding 8's rationale about the file being
copied forward, and `L1CDM2CHAIN` already exists to resolve it. Widening `category(sel)` to `category(e)`
and keying off `L1CDM2CHAIN[lc(e.target)]` closes it.

### C. There is a fourth address collision, and it is in the collisions table's blind spot — should fix

The README's ⚠ section covers `0x9338b515…` and `0x87c511c8…`. It misses one that this PR now depends on:

**`0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE`**

| chain | evidence | contract |
|---|---|---|
| Ethereum | `getMinDelay()` returns `0`; `GuardCM.owner()` returns it | the **Timelock** |
| Mode | `owner()` = `drainer()` = `0x9338b515…` (the mediator); `VERSION()` = `1.0.0`; no `serviceRegistry()` | **ServiceRegistryL2** |

It is already load-bearing: the new test defines `MODE_SERVICE_REGISTRY_L2` as that literal at
`Proposal13Housekeeping.t.sol:60` while `TIMELOCK` in the builder is the *same 20 bytes*, and
`_modeCmScheduleCall()` uses both meanings in one call — `TIMELOCK` as the guard's `to`, and
`MODE_SERVICE_REGISTRY_L2` inside the bridged payload. Anyone applying the README's own instruction
("read before editing any constant") would not learn this from the table. Two constants with identical
values and opposite meanings, in the same file, is precisely the hazard the section exists to prevent.

Suggest adding the row to the table, a comment at the constant, and — for consistency with fix 1 — a
`34443:0x3c1ff68f…` entry in `ADDR_BY_CHAIN`, so a future proposal that annotates a Mode `drain()` does
not print "Timelock".

---

### Minor

- **The audit row is intentionally ahead of on-chain state.** Until the proposal executes, the Mode row
  logs three mismatches. Not a break — `customExpect` catches and logs rather than throwing, and it does
  not set the non-zero exit flag — but a line saying the row is expected to fail until execution would save
  the next person triaging that output.
- **The new drift guard is advisory only.** `.github/workflows/workflow.yaml` runs eslint → solhint →
  `hardhat test` → coverage; no `forge` step. So `test_committedArtifactsMatchTheBuilder` — the thing that
  catches `calldata.json` drifting from the builder — only fires when someone runs Forge by hand. Nothing
  to fix in this PR, but it means the guard is documentation-plus-a-manual-check rather than enforcement.
- **The volume figures** — I tried again against four public Polygon endpoints and hit the same
  `eth_getLogs` range caps, so they remain un-reverified on my side. Rather than re-running it, the cheaper
  fix is to record the block range and endpoint used in the README next to the numbers, so the claim is
  checkable by anyone without re-deriving the scan.

Nothing above touches the calldata, so none of it blocks submission on correctness grounds — A and C are
about the tests and docs holding their value, and B is about the next proposal that copies this file.

DavidMinarsch:
All three actioned, plus the three minors. **Proposal inputs untouched again** — `calldata.json`, `description.txt` and the builder are byte-identical, so the pinned identifiers cannot have moved. 10/10 pass.

## A — you caught finding 6 reappearing in the one test where it matters

Correct, and embarrassing: I tightened the L2 negative tests for exactly this reason **in the same commit** that left the capability test's before-leg bare. The whole weight of "rejected before, accepted after" rests on that leg failing for the stated reason, and a bare `expectRevert` would have gone green on a malformed schedule call or a wrong `cm` — the "before" half proving nothing.

Pinned:

```solidity
vm.expectRevert(
    abi.encodeWithSignature(
        "NotAuthorized(address,bytes4,uint256)", MODE_L1CDM, bytes4(0x3dbb202b), uint256(1)
    )
);
```

Your point about the value being more than pedantic is the better half: `chainId 1` + `sendMessage` **is** the direct-L1 delta the README describes, so the pin now documents the pre-state rather than just asserting a failure. Said so at the call site.

## B — right, and it is the same defect one function up

`0x3dbb202b` is `sendMessage` on *any* OP-stack messenger. `category()` now resolves through `L1CDM2CHAIN[lc(e.target)]` exactly as `addrSpan` does, and I added `optimism`/`base`/`celo` headings so a copy-forward carrying one of those legs lands somewhere correct rather than under a Mode-titled heading.

## C — a fourth collision, and it was inside my own test

Verified on the chain each meaning is used on:

| chain | evidence | contract |
|---|---|---|
| Ethereum | `getMinDelay()` → `0`, no `owner()` | Timelock |
| Mode | `owner()` = `drainer()` = `0x9338b515…` | `ServiceRegistryL2` |

`_modeCmScheduleCall()` uses **both meanings in one call** — `TIMELOCK` as the guard's `to`, `MODE_SERVICE_REGISTRY_L2` inside the bridged payload, the same 20 bytes twice. Now in the README table, in `ADDR_BY_CHAIN` for **both** chains (so a Mode `drain()` annotation cannot print "Timelock"), and as a comment at the constant saying explicitly not to collapse the two.

That the collisions section missed the collision sitting in the code it introduced is the strongest argument yet for the section existing.

## Minors

- **Audit row ahead of state** — README now says the Mode row logs three mismatches until execution, that `customExpect` logs rather than throws, and that the run still exits 0. So nobody triages it as a regression.
- **Drift guard advisory only** — agreed, and left alone. Adding a `forge` step to CI is a repo-wide change, not a proposal PR's business.
- **Volume figures** — settled rather than re-run. The README now carries the method: a single Etherscan v2 `logs/getLogs` call, no chunking, reproducing **259 events, blocks `81298626`–`92060820`, 2026-01-06 → 2026-08-15**, with the per-month split. Two independent attempts hit range caps, which is the same empty-versus-refused ambiguity that made the `olas-lst` base leg wrong — so the note also says to count unreadable chunks if a chunked scan is used instead.

kupermind:
## Re-review — 9213ef8

A, B, C and the three minors are all addressed and verified. **Proposal inputs are unchanged since the
first commit** — `calldata.json`, `description.txt` and `Proposal13Housekeeping.s.sol` are byte-identical
across all three commits, so the pinned identifiers have not moved at any point. 10/10 pass; eslint and
solhint 0 errors.

| # | Fix | How I checked it |
|---|---|---|
| A | Pinned `NotAuthorized(MODE_L1CDM, 0x3dbb202b, 1)` | Argument types match the declaration at `VerifyData.sol:8` exactly; the test passes with the pin, so the live revert matches it byte-for-byte |
| B | `category(e)` resolves via `L1CDM2CHAIN[lc(e.target)]`, plus `optimism`/`base`/`celo` headings | Grouping for entry 1 is unchanged (`0x95bDCA6c…` → 34443 → `mode`), and the empty new headings are skipped by `if (!idx \|\| !idx.length) continue;` rather than emitted |
| C | Collision in the README table, in `ADDR_BY_CHAIN` for both chains, and commented at the constant | Re-confirmed the asymmetry on-chain: mainnet `0x3C1fF68f…`.`owner()` **reverts** (`execution reverted, data: "0x"`) while `getMinDelay()` answers; Mode `owner()` = `drainer()` = `0x9338b515…` |
| — | HTML reproducibility | `annotate.js` changed but `proposal_13.html` was **not** regenerated in this commit — worth checking, and it still regenerates **byte-identically**. The new `ADDR_BY_CHAIN` rows and the `category()` rewrite are genuinely no-ops for this proposal's rendering |

Adding the `1:0x3c1ff68f…` row alongside the Mode one was the right instinct — pinning both halves of a
collision is what stops the flat map from silently answering for the wrong chain later.

---

### 1. The `<details>` block eats the paragraph after it — should fix

`README.md:90` closes the block and resumes prose on the same line:

```markdown
</details> After execution, a service that would have deployed a PolySafe
deploys an ordinary Safe instead. `GnosisSafeMultisig`
([`0x3d77596b…`](https://polygonscan.com/address/0x3d77596beb0f130a4415df3D2D8232B3d3D31e44)) remains
```

In GFM an HTML block runs to the next blank line and its contents are not parsed as Markdown, so
everything up to `reverts in the meantime.` is emitted raw. Confirmed against GitHub's own renderer
(`POST /markdown`) — the backticks and the link stay literal:

```html
 After execution, a service that would have deployed a PolySafe
deploys an ordinary Safe instead. `GnosisSafeMultisig`
([`0x3d77596b…`](<a href="https://polygonscan.com/address/0x3d77596b…">https://polygonscan.com/address/0x3d77596b…</a>)) remains
```

No paragraph break, visible backticks, and the `GnosisSafeMultisig` link renders as literal
`([`0x3d77596b…`](…))` with only the bare URL autolinked. A blank line after `</details>` fixes it — same
renderer, same input:

```html
<p>After execution, … <code>GnosisSafeMultisig</code>
(<a href="https://polygonscan.com/address/0x3d77596b…"><code>0x3d77596b…</code></a>) remains …</p>
```

Worth fixing because the paragraph it mangles is the fallback-multisig statement — the part of the entry-2
write-up a reviewer most needs to read cleanly.

### 2. The documented `getLogs` call has no topic filter — should fix

The prose says the creator emitted 259 **`MultisigCreated`** events, but the recipe filters on address
only:

```
&address=0xA749f605…&fromBlock=0&toBlock=latest
```

That returns *every* log the creator has emitted. The two counts coincide only if `MultisigCreated` is the
sole event on that contract — which may well be true, but the note exists precisely so the figure does not
have to be taken on trust, and as written it does not reproduce the claim it is attached to. Adding
`&topic0=<MultisigCreated topic>` closes the gap and makes the command match the sentence.

What I could check without a key is internal consistency, and that holds: the monthly split sums exactly
to the headline figure (`8+63+75+92+4+10+4+3 = 259`), and the `92 → 3` in the prose matches `04=92`,
`08=3`. The address in the URL is also the correct EIP-55 checksum of the builder's `POLYSAFE_CREATOR`.

I still cannot verify the figures themselves — no Etherscan key here, and two more public-RPC attempts
timed out. That is fine; recording the method was the ask, and it is now recorded.

### 3. One caveat worth a line in that note — nit

Etherscan's `logs/getLogs` paginates at 1000 records per call. At 259 the single call is correct, but the
note frames it as the general recipe ("no chunking required"), and past 1000 it returns a full-looking
page rather than an error. That is the same silent-truncation failure the paragraph below it warns about,
so it would be consistent to say: single call is valid while the count stays under 1000, else paginate
with `page`/`offset`. (Worth confirming the current v2 limit before wording it.)

---

Nothing here touches the calldata, and none of it changes my read of the proposal: the three entries encode
what the description says, the identifiers are stable across all three commits, and the capability entry 0
exists for is now asserted end-to-end with a pinned pre-state. Items 1–3 are about the write-up reproducing
its own claims.

DavidMinarsch:
Both fixed and verified the same way you checked them.

## 1 — `</details>` swallowing the paragraph

Confirmed and fixed. Verified through GitHub's `/markdown` endpoint before and after; it now renders as a real paragraph:

```html
<p>After execution, … <code>GnosisSafeMultisig</code>
(<a href="https://polygonscan.com/address/0x3d77596b…"><code>0x3d77596b…</code></a>) remains …</p>
```

No literal backticks, and the polygonscan link is an actual anchor rather than a bare autolink.

Your point about *which* paragraph it was mangling is the reason it mattered — the fallback-multisig statement is the load-bearing sentence of entry 2, and it was the one rendering as raw text.

## 2 — topic filter on the recipe

Right, and the distinction is worth stating plainly: the recipe exists **so the figure need not be taken on trust**, and as written it relied on a coincidence to be correct.

Added `topic0 = keccak("MultisigCreated(address,address)")` = `0x38b56a7814c1a51663dba0756bf5ece28e95bcf82cf63a76479424baf9c680f9`, and re-ran it — the filtered query returns **259**, the same figure. So the counts did coincide (`MultisigCreated` is the only event on that contract), but the command now proves the sentence instead of matching it by luck.

Your internal-consistency check is a good one to have recorded: the monthly split summing to 259 and the `92 → 3` matching `04=92`/`08=3` is exactly the kind of thing that survives without an API key.

## On the HTML

Confirmed your observation — `annotate.js` changed without `proposal_13.html` being regenerated, and re-running the documented command produces a **byte-identical** file. So the `ADDR_BY_CHAIN` rows and the `category()` rewrite are genuine no-ops for this proposal's rendering, as you found. I have left the committed HTML as-is rather than touching its mtime for nothing.

---

Proposal inputs remain byte-identical across all four commits — `calldata.json`, `description.txt` and the builder have never moved, so the pinned `proposalId` and `descriptionHash` still hold. CI green (build, gitleaks, lint). Nothing outstanding from my side; it needs an approving review to unblock.

PR #230, opened 2026-09-04
proposal 14: sync the L2 withheld OLAS amounts back to L1

DavidMinarsch:
Proposal 14 — seven bridged actions instructing each L2 staking target dispenser to report its withheld OLAS amount back to the L1 `Dispenser`. Built to the same shape as proposals 11–13: builder script as single source of truth, `description.txt` + `calldata.json` + annotated HTML, L1 fork test through the live Governor, per-chain L2 leg tests.

**proposalId** `40126612048141097336620033593318472673469120154184180144739203820830139231953`
**descriptionHash** `0x993cbb52a409a856560c105510540d773c41589ee5a5b2e46247e5c66801f9ee` (unchanged — `description.txt` has not moved)

| # | Chain | Route | Withheld (OLAS) |
|---|---|---|---:|
| 0 | Polygon | FxRoot → FxGovernorTunnel | 100,408.03 |
| 1 | Gnosis | AMB → HomeMediator | 2,620,457.10 |
| 2 | Optimism | L1CDM → OptimismMessenger | 225,834.65 |
| 3 | Base | L1CDM → OptimismMessenger | 1,380,908.80 |
| 4 | Celo | L1CDM → OptimismMessenger | 724.80 |
| 5 | Mode | L1CDM → OptimismMessenger | 703,245.03 |
| 6 | Arbitrum | Inbox retryable → dispenser directly | 724.80 |

**No tokens move.** The reported figure nets against *future* transfers to the same chain (`Dispenser.sol:728-740`, `:1193-1207`), so the protocol stops sending OLAS to chains already holding an unspent balance. The path has never been used: every L2 `stakingBatchNonce` and every L1 `mapChainIdWithheldAmounts` entry reads `0`, against ~5.03M OLAS held. The amounts are **not parameters** — each dispenser reads its own `withheldAmount` at execution.

### Three things reviewers should look at

**Two wire formats.** Polygon carries the packed buffer directly (FxChild calls `processMessageFromRoot` itself); Gnosis and the OP-stack chains need the **encoded** `processMessageFrom…(bytes)` call wrapping it. Getting this wrong fails only on the destination chain — `requireToPassMessage` and `sendMessage` just enqueue, so the L1 leg passes either way. Every L2 leg test therefore replays the builder's own `packedFor(...)` bytes.

**Arbitrum has no governance receiver.** The dispenser's owner is the Timelock's L2 alias (`0x3C1fF68f…` + `0x1111…1111` = `0x4d30F68F…a70F`), asserted in a test rather than assumed. Entry 6 is a retryable ticket calling the dispenser directly, and the only entry carrying ETH — ≈0.0000308 ETH, which **the Timelock pays from its own balance**, so no top-up and no ETH attached at execution.

**Address collision.** `0x9338b515…` is both Polygon's `FxGovernorTunnel` (entry 0) and Mode's mediator (entry 5). Both are used here; each was verified by a chain-specific getter on the chain it is used on, and the HTML resolves chain from the L1 entrypoint.

### Testing — 20 passing

- **L1** (5, mainnet fork through the live GovernorOLAS): preconditions; full propose → vote → queue → execute, **counting** all seven dispatches rather than spot-checking, since four entries share an event signature; execute-as-Timelock; proposalId pin; artifacts-match-builder. `execute()` = 9,938,629 gas vs the 16,777,216 cap.
- **L2** (15, across seven chains): each leg replays the builder's packed buffer through its receiver and asserts the dispenser syncs to dust with the nonce advanced, plus a negative test pinning the **exact** revert.

**One environment limit, stated rather than hidden:** Arbitrum's `sendTxToL1` is the ArbSys precompile at `0x64`, which a plain EVM fork does not implement — the unmocked call dies with `InvalidFEOpcode`. The test mocks it; that the real precompile accepts the call was established by `eth_call` against a live Nitro node (succeeds from `owner()`, reverts `OwnerOnly` from anyone else), recorded in the README.

### Before submission

Regenerate entry 6 — `maxSubmissionCost` tracks the L1 base fee, so it and the proposalId will move. Re-read the withheld figures for the proposal text. And note the proposal **is not finished when it executes**: every leg leaves an L2→L1 message that must be claimed (Polygon ~30–90 min, Gnosis hours, OP-stack and Arbitrum 7 days), confirmed per chain by reading `mapChainIdWithheldAmounts` on L1.


DavidMinarsch:
Right, and it was worse than a buffer question — fixed in `aae781d`.

**The fee is recomputed at execution, not at build.** The Inbox charges `(1400 + 6 × dataLength) × block.basefee` — `1808 × basefee` for this 68-byte payload — and reverts if `maxSubmissionCost` is short. Entry 6 is in the same batch as the other six, so **that revert takes all seven entries down**, ~5 days after the vote, on a number nobody looked at again.

Now priced as a ceiling rather than a snapshot:

| | before | now |
|---|---|---|
| base fee covered | 0.41 gwei (build-time) | **45 gwei (~109x)** |
| `maxSubmissionCost` | 809,502,808,032 | 81,360,000,000,000 |
| `gasLimit × maxFeePerGas` | 300,000 × 0.1 gwei | 300,000 × 0.05 gwei |
| entry 6 value | 30,809,502,808,032 | **96,360,000,000,000 (≈0.0000964 ETH)** |

Still inside the Timelock's own 0.0001 ETH, so no top-up and nothing attached at execution.

**Why the L2 leg dropped to 0.05 gwei to pay for it:** the two shortfalls aren't symmetric. An L1 shortfall reverts the proposal; an L2 shortfall only means the ticket needs redeeming by hand within 7 days. So the budget goes to the L1 side. 0.05 gwei is still ~5x Arbitrum's 0.01 gwei floor.

`test_arbitrumEntrySurvivesABaseFeeSpike` asserts the ceiling **from both sides** — executes at 45 gwei, fails a gwei above — so it's a fact rather than a comment.

**A second bug the same review surfaced.** Both refund addresses were `TIMELOCK` — the L1 address. On Arbitrum that address belongs to nobody, so every refund (the excess submission fee, and unused L2 gas) would have been **permanently lost**. They're now the alias, which this same governance route can spend. `test_arbitrumRefundsGoToTheAlias` decodes the committed calldata and pins it.

Three things worth knowing about the headroom:

- **Over-pricing costs only float** — the excess over the actual fee is refunded on L2 to the alias.
- **A reverted execution is not a lost vote.** `execute()` can be retried and a queued proposal has no deadline, so if the base fee is above the ceiling on the day, the fallback is to execute in a calmer block.
- **The ceiling is raisable without a vote.** Sending ETH to the Timelock is permissionless, so a top-up plus a higher `ARB_MAX_SUBMISSION_COST` buys any ceiling you want. Say the word if you'd rather price it for, say, 200 gwei and top the Timelock up to 0.001 ETH.

New proposalId `40126612048141097336620033593318472673469120154184180144739203820830139231953`. 22 tests pass.

test/proposals/Proposal14L2Legs.t.sol
```diff
@@ -0,0 +1,210 @@
...
+}
+
+abstract contract Proposal14OpLegBase is Proposal14LegBase {
+    address internal constant L2_MESSENGER = 0x4200000000000000000000000000000000000007;
+
+    function _opSyncs(address messengerL2, address dispenser) internal {
+        _assertPrecondition(dispenser);
+        vm.mockCall(L2_MESSENGER, abi.encodeWithSignature("xDomainMessageSender()"), abi.encode(TIMELOCK));
+        vm.prank(L2_MESSENGER);
+        IReceiverOp(messengerL2).processMessageFromSource(packedFor(dispenser, abi.encode(RETURN_GAS)));
```
OjusWiZard: **Important — the L2 leg tests never read `buildProposal()`'s calldatas, so the two failure modes this PR names as central are still untested.**

Every leg builds its own input with `packedFor(...)` and then hand-picks the wrapper by choosing the typed interface — `IReceiverOp.processMessageFromSource` here, `IReceiverFx.processMessageFromRoot` at :66, `processMessageFromForeign` at :89. The wrapper the test uses is therefore independent of the wrapper the builder wrote, so the test cannot disagree with the builder about wire format. `buildProposal()` is never called in this file and `calldatas[i]` is never read.

The L1 side doesn't close it either: `_assertAllSevenDispatched` counts the four OP-stack entries by emitter union (`opCount == 4`) without decoding `SentMessage`'s target or message, so two entries pointed at the same L1CDM still passes; `sawGnosis` / `sawArbitrum` are bare bools.

So both errors the builder header calls out in bold — the wrong wrapper, and a mispaired (L1CDM, L2 receiver, dispenser) triple — would ship green. The README's claim that the L2 leg tests are what distinguish the two formats isn't accurate as written.

To be clear: I decoded `calldata.json` at `aae781d7` and **the wrappers and pairings are all correct today**. This is a missing safety net rather than a shipped bug.

`test_arbitrumRefundsGoToTheAlias`, added in `aae781d7`, is the pattern — it's the first test that decodes `calldatas[6]` rather than restating what the builder should have produced. Extending that to the other six would close this: decode each leg's buffer out of `calldatas[i]` (past the `sendMessage` / `requireToPassMessage` / `sendMessageToChild` args, and for the five wrapped routes past the `processMessageFrom...` selector) and feed *that* to the receiver, and assert the `SentMessage` target per entry instead of counting emitters. proposal_13 did assert the `SentMessage` target (`Proposal13Housekeeping.t.sol:169-172`); this went from one OP-stack entry to four and dropped it.

scripts/proposals/proposal_14/annotate.js
```diff
@@ -0,0 +1,425 @@
...
+        body += `<h2>${esc(name)} (${idx.length})</h2>`;
+        for (const i of idx) {
+            const e = entries[i];
+            body += `<div class="entry"><div class="ehead"><span class="ix">[${i}]</span> target = ${addrSpan(e.target)} &nbsp; value = <span class="val">${esc(e.value)}</span></div>` +
+                decodeEntry(e) +
+                `<details class="raw"><summary>raw calldata</summary><pre>${esc(e.calldata)}</pre></details></div>`;
+        }
+    }
+    const valNote = nonZero.length
+        ? `All values are 0 EXCEPT ${nonZero.map((i) => "entry [" + i + "]").join(", ")} (Arbitrum retryable), which carries a non-zero value (deposit&times;10 buffer) supplied by the executor of execute() and forwarded through the Timelock to the Inbox.`
```
OjusWiZard: **Important — this line renders text that is false for proposal 14, in the artifact voters read.** `aae781d7` made it more wrong rather than less.

`proposal_14.html:24` still reads:

> ...which carries a non-zero value **(deposit x10 buffer) supplied by the executor of `execute()`** and forwarded through the Timelock to the Inbox.

Both halves are wrong:

- **"deposit x10 buffer"** — there is no deposit and no x10. After `aae781d7` the value is deliberately priced as a *ceiling*, `1808 * 45 gwei + gasLimit * maxFeePerGas`, which the constant's own comment explains well. The rendered sentence describes neither the old pricing nor the new.
- **"supplied by the executor of `execute()`"** — contradicted by the README, by the builder header, and by `Proposal14WithheldSync.t.sol`, which executes with **no value** specifically to prove the Timelock funds it from its own balance.

Root cause: the string is verbatim from `proposal_12/annotate.js:297` and `proposal_13/annotate.js:371`, behind the `nonZero.length ?` ternary whose else-branch on the next line ("Every entry has value 0.") is what those proposals actually rendered. **Proposal 14 is the first with a non-zero value, so this branch has never executed before** — worth a look at any other dormant text in the generator for the same reason.

Since it's the one line in the artifact not derived from the calldata, deriving it would beat rewording it: print `maxSubmissionCost`, the base fee it tolerates (`ARB_BASEFEE_CEILING` is already exposed), and who funds it.

OjusWiZard (commented):
Re-reviewed at `aae781d7`. I recomputed `descriptionHash` and `proposalId` independently, rebuilt and decoded all seven entries, and re-checked the alias arithmetic, the wire format per route, the `0x9338b515...` collision and every dispenser owner — all correct at this head.

`aae781d7` resolves what I had as the blocking issue. For the record, at `d11bfa75` `ARB_MAX_SUBMISSION_COST` sat at 1.0x the build-time fee, i.e. a break-even L1 base fee of 0.4477 gwei; I measured 474/1025 recent blocks above that, so `execute()` had roughly even odds of reverting — and because `executeBatch` is atomic it would have taken all seven legs with it, not just Arbitrum. Pricing it as a 45 gwei ceiling and exposing `ARB_BASEFEE_CEILING` for the fork assertion is the right shape, and dropping the L2 leg to 0.05 gwei to stay inside the Timelock's balance is the correct trade: an L1 shortfall reverts everything, an L2 shortfall only needs a manual redeem inside 7 days. Refunds to the alias are right too — I confirmed both fields decode to `0x4d30F68F...A70F`.

Two things from the earlier pass are untouched by that commit, inline below. Neither is a defect in the calldata, which I verified byte-for-byte.

One housekeeping note: the PR description still pins the old `proposalId` `46220811...64680`. The committed artifacts and my own recomputation both give `40126612048141097336620033593318472673469120154184180144739203820830139231953` at this head (`descriptionHash` is unchanged, since `description.txt` didn't move). Worth updating the description so voters diffing the two don't stall.

OjusWiZard (dismissed):
Approving at `aae781d7`.

The blocking issue from my previous pass is resolved, and I re-verified this head from scratch: `descriptionHash`, `proposalId`, all seven `(target, value, calldata)` entries rebuilt and decoded byte-for-byte, the alias arithmetic, both wire formats per route, the `0x9338b515...` collision, and every dispenser owner against its route. The Arbitrum entry now decodes to `maxSubmissionCost + gasLimit * maxFeePerGas` with both refunds at the alias, and the value math is self-consistent.

The two inline comments still stand, but neither is a defect in the calldata and neither should hold up the merge:

- the L2 leg tests not reading `buildProposal()`'s calldatas is a missing safety net, not a shipped bug — the wrappers and pairings are correct at this head, I just can't point at a test that proves it;
- the `annotate.js` value note is wrong prose in a generated artifact with no effect on what executes.

The second one is worth clearing **before on-chain submission** rather than before merge, and it won't clear itself: regenerating the artifacts re-renders the same sentence, since the string lives in `annotate.js` and not in the calldata. The README's pre-submission regeneration step is the natural place to catch it.

Two things for that same pass: the PR description still pins the superseded `proposalId` `46220811...64680` (the artifacts and my recomputation agree on `40126612...31953`), and entry [6] now sits at ~96.4% of the Timelock's balance — fine as-is and topping up is permissionless, but there's no room for a further ceiling raise without sending ETH first.

DavidMinarsch:
Both inline comments are addressed in `f34e736`, and the PR description now pins `40126612…31953`.

## 1 — the L2 legs never read the builder's calldatas

Correct, and the sharpest part of the diagnosis is that **a test that picks its own wrapper cannot disagree with the builder about the wire format**. Choosing `IReceiverOp` over `IReceiverFx` *was* the assertion, and it was an assertion about my own test.

Every leg now takes its input from `buildProposal()` and delivers **the exact bytes the bridge would deliver**, by low-level call — the `_message` out of `sendMessage`, the `_data` out of `requireToPassMessage`, the packed buffer out of `sendMessageToChild`, the `data` out of `createRetryableTicket`. No typed interface, no re-encoding. The receiver and dispenser are read out of the calldata and then checked against that chain's constants, which closes the mispairing half.

**Proven by mutation rather than by inspection**, since that is exactly the mistake I just made:

| mutation | result |
|---|---|
| OP-stack hands the bare packed buffer to `sendMessage` | `entry has the wrong wrapper: 0xaea9ef99… != 0xd3042d2b…` |
| entry [3] pointed at Optimism's L1 messenger instead of Base's | `entry goes through the wrong L1 messenger: 0x25ace71c… != 0x866E82a6…` |

The L1 side is fixed too: `_assertAllSevenDispatched` no longer counts emitters. Each dispatch is matched against that entry's own bytes — OP-stack on `(emitter, SentMessage target, message)`, Polygon on the synced payload equalling this proposal's packed buffer, and the AMB and Inbox events required to contain their entry's payload verbatim. You were right that proposal 13 asserted the `SentMessage` target and this dropped it going from one OP-stack entry to four.

The README's claim is now accurate rather than aspirational, and it names the two mutations.

## 2 — `annotate.js` renders false text

Also right, including the root cause: the string is inherited from proposals 12 and 13, behind a ternary whose non-zero branch **had never executed** because no earlier proposal carried a value.

Rather than reword it, the note is now derived from entry [6]'s own bytes — `maxSubmissionCost`, `gasLimit × maxFeePerGas`, and the base fee it tolerates, computed from `(1400 + 6 × dataLength)` against the actual payload rather than restated:

> All values are 0 EXCEPT entry [6] (Arbitrum retryable) carries 96360000000000 wei = maxSubmissionCost 81360000000000 + gasLimit 300000 × maxFeePerGas 50000000 (15000000000000). The Inbox recomputes the submission fee at execution as (1400 + 6×dataLength) × block.basefee, so this prices an L1 base fee up to **45.0 gwei**; above that the entry reverts, and with it the whole batch. The Timelock funds it from its own balance — no value is attached to execute().

So it re-derives on every regeneration instead of needing to be caught in the pre-submission pass. Your wider point stands — that generator has other branches that have never run, and they are worth a look before the next proposal reuses it.

## Housekeeping

- PR description updated to `40126612…31953`; `descriptionHash` is unchanged, as you noted.
- **Entry [6] at ~96.4% of the Timelock's balance is deliberate and confirmed with David**: 45 gwei is ~109x the build-time base fee, over-pricing costs only float, a reverted `execute()` is retryable against a calmer block, and a top-up is permissionless if a higher ceiling is ever wanted.

22 tests pass.

OjusWiZard (approved):
Re-approving at `f34e7368`. Both inline comments are addressed and I've resolved them.

**The L2 legs now take their input from the builder.** `_delivered(i)` calls `buildProposal()`, strips the selector and decodes `calldatas[i]` per route — `(address,bytes,uint32)` for the OP-stack, `(address,bytes,uint256)` for Gnosis, the nested FxChild wrapper for Polygon, the 8-tuple for Arbitrum — then unwraps the `processMessageFrom...` layer to recover the packed buffer. That closes the gap: the wrapper the test delivers is now the wrapper the builder wrote, so a raw-vs-wrapped mix-up can no longer pass.

`_assertAllSevenDispatched` got the better half of the fix. The four OP-stack entries are matched per entry on `(emitter, target, message)` rather than counted by emitter union, so the mispaired `(L1 messenger, L2 receiver)` case I described survives nothing; Polygon asserts the synced payload equals the proposal's own buffer, and Arbitrum is now pinned to `ARBITRUM_INBOX` with the L2 calldata asserted inside the ticket.

**The value note is derived rather than asserted.** `describeValue()` decodes entry [6] and prints the actual split — `96360000000000 wei = maxSubmissionCost 81360000000000 + gasLimit 300000 x maxFeePerGas 50000000` — computes the tolerated base fee from the entry's own `dataLength` (45.0 gwei, which matches my own arithmetic), and states both that exceeding it reverts the whole batch and that the Timelock funds it with nothing attached to `execute()`. All four claims are now true and all four come out of the bytes.

Worth recording that **the on-chain payload did not move**: `calldata.json`, `description.txt` and `Proposal14WithheldSync.s.sol` are byte-identical to `aae781d7`, so `proposalId` `40126612...31953` and `descriptionHash` `0x993cbb52...f9ee` still hold, and the entry decode I verified at that head still describes exactly what would execute. This commit changed only the tests and the annotator.

One advisory is still open and I'm not holding the merge for it: the Arbitrum negative test (`Proposal14L2Legs.t.sol:296`) still uses `assertFalse(ok, ...)` and accepts any revert, while its six siblings now pin the exact custom error. The README records the expected `OwnerOnly(...)` from a live call, so it's a one-line change whenever you're next in the file.

PR #231, opened 2026-09-06
feat(guardcm): Robinhood Chain (4663) bridge mediator params

DavidMinarsch:
Closes **D3** of the Robinhood Chain rollout audit (plan: valory-xyz/autonolas-analytics#254).

§4 entry 1 of the rollout needs `GuardCM.setBridgeMediatorL1BridgeParams` for chain 4663, but `script_26_03_set_bridge_mediators_arbitrum.sh` **hardcodes `arbitrumChainId=42161`** (line 54) and `arbitrumInboxAddress` points at Arbitrum One's inbox, so it cannot be reused as-is.

### What's here

- `script_26_05_set_bridge_mediators_robinhood.sh`, generated from the Arbitrum script by key renaming and the chain Id, so the two stay in lockstep.
- Two globals keys: `robinhoodInboxAddress` (Robinhood Chain's Delayed Inbox on Ethereum, `0x1A07cc4B…`) and `robinhoodBridgeMediatorL2`.

`ProcessBridgedDataArbitrum` is **reused unchanged** as the L2 verifier — it is chain-agnostic and Robinhood Chain is an Arbitrum Orbit rollup, so no new verifier contract is needed.

### One value will look wrong on review — it isn't

`robinhoodBridgeMediatorL2` is **byte-identical** to `arbitrumBridgeMediatorL2` (`0x4d30F68F…A70F`). That is correct: both are the L2 alias of the *same* L1 Timelock, and address aliasing is chain-independent (`Timelock + 0x1111…1111 mod 2^160`). **Only the inbox and the chain Id actually differ** between the two chains.

### Why this script exists at all

Mode had its four GuardCM allowlist entries backfilled in July 2026 but `setBridgeMediatorL1BridgeParams` was never called, so `verifierL2` is zero and community-multisig transactions to Mode **fail closed** — the one supported L2 the CM cannot pause. That repair is proposal 13 entry 0, currently `Queued`. Shipping the allowlist without the bridge params is precisely the failure this avoids repeating for 4663.

No contract changes. Script and config only.


DavidMinarsch:
Closing this — it's the wrong shape, and the review question that prompted it was right.

`GuardCM.setBridgeMediatorL1BridgeParams` is owner-gated, and `GuardCM.owner()` is the **Timelock** `0x3C1fF68f…` (verified on-chain). So `cast send` from a Ledger or `PRIVATE_KEY` reverts `OwnerOnly` — this call can only be made **through a governance vote**.

That means I misread the script I generated this from. `script_26_03_set_bridge_mediators_arbitrum.sh` and its siblings are **deployment-era artifacts**, added in `ef2531f` ("deployment scripts for newly deployed contracts") and runnable only while the deployer still owned GuardCM. They are not a live operational route, so cloning one for Robinhood produced a script that cannot work.

The repo already has the right mechanism: `scripts/proposals/proposal_N/` — a Forge builder emitting the authoritative calldata, `calldata.json`, `annotate.js` → an HTML breakdown, and a `description.txt` that must match the builder's `DESCRIPTION` byte-for-byte. **Proposal 13 entry 0 is literally this call** for Mode (target `0xC0b146D6…`, selector `0x1602c55c`), so the Robinhood entry is the same shape, differing only in the L1 entrypoint (Robinhood's Delayed Inbox rather than the Mode L1CDM), the verifier (`ProcessBridgedDataArbitrum` rather than `…Optimism`), the chain Id, and the mediator.

The Robinhood entries will ride a single proposal covering the whole rollout — the `Dispenser` wiring, the GuardCM bridge params and the GuardCM selectors together — so it belongs in a proposal folder built once wave 1 has deployed and the L1 processor address exists. Nothing in this PR survives that: the two globals keys were only ever inputs to the script, and the proposal builders hardcode their constants.

One thing worth carrying into that proposal: `proposal_13`'s builder opens with an address-collision warning, because three unrelated contracts share `0x9338b515…` through aligned deployer nonces. Robinhood has the same trap from a different cause — its L2 mediator is byte-identical to Arbitrum One's, since both are the alias of the same L1 Timelock. Correct, but it will look wrong on review, so it wants the same kind of note.

Tracked in the rollout plan at valory-xyz/autonolas-analytics#254 §4.

PR #232, opened 2026-09-06
proposal 15: raise the Governor's proposal threshold and quorum

DavidMinarsch:
Two actions, both on the Governor itself, both `onlyGovernance` (executor is Timelock A). No value, no bridge, no external contract.

| # | Call | From | To |
|---|---|---:|---:|
| 0 | `setProposalThreshold(uint256)` | 5,000e18 | **250,000e18** |
| 1 | `updateQuorumNumerator(uint256)` | 3 | **10** |

**proposalId** `3563000702947626407249966116394330391891319244018918786718598572660765109424`
**descriptionHash** `0xbc86418e14715cfad78c62576109d9825e2b1b7395fb5f3a594cc64eef35f8b1`

## Why

A proposal created 2026-09-05 asked the DAO to transfer Treasury ownership to an externally owned account. Its proposer held 5,100 OLAS — just over the 5,000 veOLAS minimum, which has not been revisited since deployment.

**250,000 is the top of a flat band.** The five largest positions are all above it and the sixth is below 16,000, so every threshold in between admits the same five addresses. It is still a real restriction against today's setting — twelve addresses clear 5,000, five clear 250,000 — and the description states that rather than eliding it; holders in between regain eligibility by increasing or extending a lock.

**Quorum matters more than the threshold.** Bravo counts only For votes toward quorum, so quorum is the floor an *unopposed* proposal must reach. Every proposal in this Governor's history was carried by a single voter fielding 286,101–1,126,974 votes, and at a 10% numerator each of the four largest positions still clears quorum alone. 15% would need two voters to coordinate, which has never happened.

## Safety

- **250,000 stays below the power of the address that creates the DAO's proposals** — `cancel(proposalId)` is permissionless once a proposer falls under the threshold, so a higher bar would make the DAO's own live proposals cancellable by anyone. Asserted in the test; worth re-checking before execution, as veOLAS power decays.
- **Nothing in flight is affected.** `quorum(blockNumber)` reads a checkpointed numerator, so proposals already snapshotted keep 3%. The new numerator binds at the execution block.
- Neither change defends the Treasury against a funded proposer; both raise the cost of putting a proposal in front of the DAO that it must then mobilise to defeat.

## Tests

```
ETH_RPC_URL=<rpc> forge test --match-contract Proposal15 -vv
```

3/3 passing against live mainnet state: builder bytes + published id and hash; the description ends with the DAO Constitution reference; and the full lifecycle — propose, vote, queue, execute — checking `proposalThreshold`, `quorumNumerator`, `proposalEta == queue + governorDelay`, and that a snapshotted proposal keeps 3%. The lifecycle runs on a rehearsal description so the suite keeps working once the real proposal is on-chain; the published bytes are asserted view-only.

Analysis behind the parameter choice is in `autonolas-analytics` under `governance/`.


scripts/proposals/proposal_15/Proposal15GovernorParams.s.sol
```diff
@@ -0,0 +1,108 @@
...
+// Timelock A, so a governance proposal is the only route to either.
+//
+//   [0] setProposalThreshold(250_000e18)  — from 5_000e18
+//   [1] updateQuorumNumerator(10)         — from 3, i.e. 3% -> 10% of voting power at a snapshot
+//
+// WHY. A proposal created on 2026-09-05 asked the DAO to transfer Treasury ownership to an
+// externally owned account. Its proposer held 5,100 OLAS, just over the 5,000 veOLAS minimum,
+// which has not been revisited since deployment.
+//
+// THRESHOLD — 250,000 IS THE TOP OF A FLAT BAND. Live voting power is concentrated: the five
```
OjusWiZard, line 19: **"250,000 is the top of a flat band" is arithmetically wrong, and the rule as stated would pick a different number.**

`Governor.propose` admits on `getVotes(...) >= proposalThreshold()` (checked against this repo's own vendored flatten, `audits/internal5/analysis/contracts/GovernorOLAS-flatten.sol:3467`). With positions 982,733 / 856,538 / 838,898 / 409,732 / 293,929 and the sixth at 15,880, the set of thresholds admitting exactly those five is the half-open interval **(15,880 … 293,929]**.

The top of that band is **293,929** — the fifth-largest position — not 250,000, which sits ~15% below it.

Both sub-claims are fine: "the five largest are all above 250,000" (min 293,929 ✓) and "the sixth is below 16,000" (15,880 ✓). It is only the "top of the band" framing that fails, and it fails self-defeatingly — by its own criterion, "taking the top of the band maximises the cost of an attempt" selects **293,929e18**, which admits the identical five addresses at a strictly higher cost. The trailing clause "without excluding anyone that a lower level inside the band would admit" is true of every point in the band, so it doesn't distinguish 250,000 either.

There is a good reason for 250,000 that neither document gives: headroom against the continuous veOLAS decay this same comment warns about eleven lines down, plus the Bravo-cancel cap. Suggest saying that instead — e.g. "250,000 sits inside a flat band whose top is ~294,000; the gap is deliberate decay headroom."

Same text, same defect, at `README.md:25-28`. Worth noting the on-chain `description` is clean — the band language was correctly scrubbed from it — so this is confined to the repo artifacts, not the immutable proposal text.

scripts/proposals/proposal_15/README.md
```diff
@@ -0,0 +1,70 @@
...
+**Threshold — 250,000 is the top of a flat band.** Live voting power is concentrated: the five largest
+positions are all above 250,000 veOLAS and the sixth is below 16,000, so every threshold between those two
+levels admits the same five addresses. Taking the top of the band maximises the cost of an attempt without
+excluding anyone a lower level inside the band would admit. Against *today's* setting it is a real
+restriction, and the description says so: twelve addresses clear 5,000, five clear 250,000, and holders in
+between regain eligibility by increasing or extending a lock.
+
+**Quorum — why 10.** Bravo counts only For votes toward quorum (`COUNTING_MODE()` is
+`support=bravo&quorum=bravo`; `_quorumReached` is `quorum(snapshot) <= forVotes`), so quorum is the floor an
+*unopposed* proposal must reach — the threshold is only the entry fee. Every proposal in this Governor's
```
OjusWiZard, line 34: **This paragraph cites turnout that undercuts the change it is arguing for.**

The "four largest positions" half is true: at `P` = 3,476,854 the new quorum is **347,685**, and 982,733 / 856,538 / 838,898 / 409,732 all clear it alone.

But the turnout range in the same sentence does not support the conclusion a reader will draw from it. The five non-zero votes ever cast on this Governor are 286,101 / 288,933 / 294,537 / 859,475 / 1,126,974 — **three of them fall below 347,685**, and they are the three lowest. Against mid-2026's larger `P` a 10% bar would have been ~419,000, so those three miss by more.

That also weakens the next sentence. "15% would require two voters to coordinate, which has never happened" is offered as the reason to stop at 10 — but on this same data, **10% would have required a second voter for three of the six past proposals**. The gap between 10 and 15 is smaller than the argument implies.

Suggest either stating the 347,685 figure and acknowledging that the three lowest-turnout proposals fall under it, or dropping the turnout range and resting the case on the position-capacity claim alone, which stands on its own.

scripts/proposals/proposal_15/calldata.json
```diff
@@ -0,0 +1,14 @@
+[
```
OjusWiZard, line 1: **`annotate.js` and `proposal_15.html` are missing — the folder does not meet the stated convention.**

`scripts/proposals/README.md:5-9` makes an "annotated, self-contained **HTML** breakdown (selectors, encoded arguments and addresses decoded and labelled, with the copy-paste `propose()` arrays and the pre-computed `proposalId`)" a required member of every `proposal_<N>/` folder.

Every prior proposal ships both:

| folder | annotate.js | .html |
|---|---|---|
| proposal_11 | ✅ | ✅ |
| proposal_12 | ✅ | ✅ |
| proposal_13 | ✅ | ✅ |
| proposal_14 | ✅ | ✅ |
| proposal_15 | ❌ | ❌ |

This file is already in the schema `annotate.js` consumes, so it looks like the generator step was simply not run. This is new drift rather than pre-existing — and the missing artifact is the one a submitter copy-pastes the `propose()` arrays from, and the one a reviewer reads to check selectors and decoded arguments without compiling. Proposals 13 and 14 also list these in a Files table in their README.

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        assertEq(v[0], 0);
+        assertEq(v[1], 0);
+        assertEq(c[0], abi.encodeWithSignature("setProposalThreshold(uint256)", uint256(250_000e18)));
+        assertEq(c[1], abi.encodeWithSignature("updateQuorumNumerator(uint256)", uint256(10)));
+
+        assertEq(keccak256(bytes(d)), PUBLISHED_DESCRIPTION_HASH, "descriptionHash");
+        assertEq(
+            uint256(keccak256(abi.encode(t, v, c, keccak256(bytes(d))))), PUBLISHED_ID, "proposalId"
+        );
+    }
```
OjusWiZard, line 93: **No `test_committedArtifactsMatchTheBuilder` — `description.txt` and `calldata.json` are unvalidated.**

Both `test/proposals/Proposal13Housekeeping.t.sol:309-350` and `test/proposals/Proposal14WithheldSync.t.sol:305-334` assert the committed files byte-match the builder, pinning count, index, target, value and calldata. `foundry.toml:14-15` grants `fs_permissions` specifically for it:

> Read-only, scoped to the proposal artifacts: the proposal tests assert that description.txt and calldata.json still match the Forge builder, which is the drift the builders warn about.

That is not true for this proposal. And proposal 14's own NatSpec names precisely the gap this test has:

> The pin above only checks the builder against itself. This checks the files the HTML and the on-chain submission are actually made from.

`testBuilderBytesAndPublishedId` does only the thing that comment calls insufficient — it pins the builder against two hardcoded constants and never reads the files.

I verified they match today (591 bytes, no trailing newline, byte-identical to the builder), so this is a missing regression guard rather than a live bug. But `description.txt` is what a human copy-pastes into `propose()`, and a single stray trailing newline there changes the proposalId with nothing to catch it — Forge does not run in CI, so there is no second line of defence.

Also missing: proposals 12, 13 and 14 all carry an explicit "description.txt MUST match the DESCRIPTION string below byte-for-byte before submission" warning in the builder. `Proposal15GovernorParams.s.sol` has none.

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        assertEq(t[0], GOVERNOR);
+        assertEq(t[1], GOVERNOR);
+        assertEq(v[0], 0);
+        assertEq(v[1], 0);
+        assertEq(c[0], abi.encodeWithSignature("setProposalThreshold(uint256)", uint256(250_000e18)));
+        assertEq(c[1], abi.encodeWithSignature("updateQuorumNumerator(uint256)", uint256(10)));
+
+        assertEq(keccak256(bytes(d)), PUBLISHED_DESCRIPTION_HASH, "descriptionHash");
+        assertEq(
+            uint256(keccak256(abi.encode(t, v, c, keccak256(bytes(d))))), PUBLISHED_ID, "proposalId"
```
OjusWiZard, line 91: **The published id is only ever checked against a copy of the builder's own arithmetic.**

This recomputes `uint256(keccak256(abi.encode(t, v, c, keccak256(bytes(d)))))` — character-for-character what the builder does at `Proposal15GovernorParams.s.sol:104-106`. Both sides of the `assertEq` are the test's own arithmetic against a hardcoded constant, so it pins the constant but cannot detect a disagreement with the Governor's own derivation.

`setUp` already selects a mainnet fork at line 68, so this is free:

```solidity
assertEq(gov.hashProposal(t, v, c, keccak256(bytes(d))), PUBLISHED_ID, "id per the deployed Governor");
```

That binds `PUBLISHED_ID` — the number in `proposal_15/README.md:7-8` and the number the DAO uses to find the proposal — to the bytecode at `0x060D0CBd…51E6`. In the local path `GOVERNOR` appears only as an array *element*, so a stale or wrong Governor constant still produces a self-consistent pass while the real Governor assigns a different id.

The companion suite in `autonolas-analytics` asserted exactly this; it looks dropped in the port. Precedent also at `test/proposals/Proposal11Activation.t.sol:138`.

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        external
+        returns (uint256);
+    function castVote(uint256 proposalId, uint8 support) external returns (uint256);
+    function cancel(uint256 proposalId) external;
+    function queue(address[] memory t, uint256[] memory v, bytes[] memory c, bytes32 dh) external returns (uint256);
+    function execute(address[] memory t, uint256[] memory v, bytes[] memory c, bytes32 dh)
+        external
+        payable
+        returns (uint256);
+    function hashProposal(address[] memory t, uint256[] memory v, bytes[] memory c, bytes32 dh)
```
OjusWiZard, line 26: Confirmed unused: `hashProposal` (declared here) and `cancel` (line 20) have zero call sites in this file. `console2` is imported at line 4 and never used either.

The two dead `IGovernor` members are worth reading as a map of the coverage that was intended and dropped — `hashProposal` for the on-chain id check, `cancel` for the safety argument. See the separate notes on lines 91 and 119.

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+
+    function testLifecycleAndPostConditions() public {
+        (address[] memory t, uint256[] memory v, bytes[] memory c,) = builder.buildProposal();
+        bytes32 dh = keccak256(bytes(REHEARSAL));
+
+        assertEq(gov.proposalThreshold(), 5_000e18, "starting threshold");
+        assertEq(gov.quorumNumerator(), 3, "starting numerator");
+        // The cap that makes this safe: the new threshold must stay below the power of the address that
+        // creates the DAO's proposals, or Bravo's permissionless cancel opens on the DAO's own proposals.
+        assertGt(gov.getVotes(DAO_PROPOSER, block.number - 1), 250_000e18, "250k stays under the DAO proposer");
```
OjusWiZard, line 119: **The safety argument is asserted as one view call on a decaying quantity, and never behaviourally.**

`README.md:39-44` ("The cap that makes this safe") reduces to this single `assertGt`. There is no test that the DAO's own live proposals do **not** become cancellable after execution, and none that the 2026-09-05 hostile proposal **does** — `cancel` is declared in the interface (line 20) and never called. The companion suite in `autonolas-analytics` tested both directions.

Two more specific problems with this line:

**It is asserted at the wrong moment.** Here `proposalThreshold()` is still 5,000e18; the lifecycle then advances `votingDelay + votingPeriod + governorDelay` — about 6.4 days of simulated veOLAS decay — before the 250,000 value actually binds. Post-execution, at lines 141-142, is the moment the invariant matters and the moment `README.md:44` calls "worth re-checking". The test has the state there and doesn't assert it.

**`proposalThreshold` is not checkpointed** (unlike `quorumNumerator`), and Bravo's cancel gate reads `getVotes(proposer, block.number - 1) < proposalThreshold()` live at cancel time. So the README section heading "Nothing in flight is affected" is true only for quorum — real in-flight proposals *are* affected through the cancel gate. `0x3447…0039`'s lock ends 2027-01-07 and its power decays linearly to zero, so this assertion is guaranteed to invert at some point before then; the hardcoded `250_000e18` records no margin. Asserting headroom (say `> 2 * 250_000e18`) would fail while there is still time to react rather than at the instant the cap is already breached.

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        assertEq(gov.state(id), QUEUED);
+        assertEq(gov.proposalEta(id), block.timestamp + gov.governorDelay(), "eta is queue + governorDelay");
+
+        vm.warp(gov.proposalEta(id) + 1);
+        vm.roll(block.number + gov.governorDelay() / 12 + 1);
+        gov.execute(t, v, c, dh);
+
+        assertEq(gov.state(id), EXECUTED);
+        assertEq(gov.proposalThreshold(), 250_000e18, "threshold raised");
+        assertEq(gov.quorumNumerator(), 10, "numerator raised");
```
OjusWiZard, line 142: **Nothing asserts the DAO can still govern after the change.**

The `WHALE` NatSpec at line 45 claims it "clears both the live and the proposed threshold, and quorum alone" — none of that is checked. The lifecycle proposes at line 122 *before* execution, so it only exercises the old 5,000 bar, and line 127 compares the cast weight against the old 3% quorum. The README's central justification for numerator = 10 — "each of the four largest positions still clears quorum on its own" — has no assertion anywhere in the suite.

This is the one failure mode that isn't recoverable: if 10% of `getPastTotalSupply` turned out to exceed every individual position, governance bricks, and the only route back is a proposal that must clear the bar that broke it. Two lines here would close it, using the state the test already has:

```solidity
assertGe(gov.getVotes(WHALE, block.number - 1), gov.proposalThreshold(), "whale can still propose");
assertGe(gov.getVotes(WHALE, block.number - 1), gov.quorum(block.number - 1), "whale still clears the new quorum");
```

test/proposals/Proposal15GovernorParams.t.sol
```diff
@@ -0,0 +1,149 @@
...
+    uint8 internal constant EXECUTED = 7;
+    uint8 internal constant FOR = 1;
+
+    Proposal15GovernorParams internal builder;
+    IGovernor internal gov = IGovernor(GOVERNOR);
+
+    function setUp() public {
+        // Fork first, then deploy the builder into it: creating a fork later would discard a builder
+        // deployed against the pre-fork state and every call to it would revert.
+        vm.createSelectFork(vm.envString("ETH_RPC_URL"));
```
OjusWiZard, line 68: **Fork configuration diverges from every other proposal test and breaks the documented invocation.**

Every other proposal test uses `vm.envOr` with a public default:

```solidity
vm.createSelectFork(vm.envOr("ETH_RPC", string("https://ethereum-rpc.publicnode.com")));
```

— `Proposal12Unnominate.t.sol:86`, `Proposal13Housekeeping.t.sol:76`, `Proposal14WithheldSync.t.sol:60`, and the L2 suites with their own per-chain vars. This file alone uses `vm.envString("ETH_RPC_URL")`: a different variable name, and no default, so it reverts when unset.

Two concrete consequences:

- `forge test --fork-url <MAINNET_RPC> -vvv`, the invocation documented at `CLAUDE.md:40`, fails `setUp` for this contract — `--fork-url` does not populate `ETH_RPC_URL`.
- Running the proposals directory with the repo's own `ETH_RPC` set fails all three tests here with a bare env-var revert while the others pass.

`Proposal11Activation.t.sol:46` also carries `assertEq(block.chainid, 1, "Must run on a mainnet fork")`, which this lacks — worth adding, since every address in the file is mainnet-specific.

OjusWiZard (approved):
## Approve

The on-chain payload is correct and fully reproducible — I recomputed every hash and byte independently and everything checks out. Findings below are all in the shipped artifact set, the rationale prose, and test coverage; none of them is a reason to hold the payload. Flagging them as Important because two of the rationale claims are things the DAO would be voting on, and because the artifact set is missing a piece every prior proposal shipped.

### Verified correct

- **`description.txt` is byte-identical to the builder's `abi.encodePacked` output** — 591 bytes, no trailing newline — **and to the `DESCRIPTION` constant in the companion `autonolas-analytics` fork test.** All three agree.
- `calldata.json` decodes to exactly `setProposalThreshold(250_000e18)` (`0xece40cc1…68400000`) and `updateQuorumNumerator(10)` (`0x06f3f9e6…000a`) against `0x060D0CBd…51E6`, values zero.
- `descriptionHash` = `0xbc86418e14715cfad78c62576109d9825e2b1b7395fb5f3a594cc64eef35f8b1` ✓
- `proposalId` = `3563000702947626407249966116394330391891319244018918786718598572660765109424` ✓, and `README.md:8`'s hex is the correct zero-padded 32-byte form.
- Every mechanism claim in the builder's header block checks out against this repo's own vendored OZ 4.8.3 flatten (`audits/internal5/analysis/contracts/GovernorOLAS-flatten.sol`): `_quorumReached` is `quorum(proposalSnapshot(id)) <= details.forVotes` (4468-4471), `COUNTING_MODE()` returns `support=bravo&quorum=bravo` (4220), the Bravo cancel guard is as described (4314), and the checkpoint claim is right — `_updateQuorumNumerator` pushes with key `clock()` and the lookup is inclusive, so the new numerator does bind *at* the execution block.
- `values` is never assigned, but `new uint256[](2)` zero-initialises, so the zeros feeding the id are guaranteed; the test pins them anyway.
- `proposalEta == queue + governorDelay` at `:134` is correct against this repo's forked `contracts/utils/GovernorTimelockControl.sol:113-116`, and appears to be the only test in the repo pinning that divergence from upstream OZ. Good assertion to have.
- The tail-comparison loop in `testDescriptionCarriesTheDaoSignature` has no off-by-one — `i ∈ [0, sig.length)` maps to exactly the last `sig.length` bytes, and the `assertGt` on line 103 guards the underflow.

I could not compile or run anything — no `forge` or `solc` in my environment — so the tests are reasoned about, not executed.

### One Important finding with no line in this diff

**The proposals index is not updated, and is now two proposals stale.** `scripts/proposals/README.md:21-25` lists 11, 12 and 13 only. Proposal 13 updated the table in the same commit that added its folder; proposal 14 did not, so that missing row is pre-existing — but this PR repeats it and touches no file outside `proposal_15/` and `test/proposals/`. The table is the documented entry point to the folders, and its "Fork test (L1)" column should point at `test/proposals/Proposal15GovernorParams.t.sol`. Worth adding both rows while you're here.

### Advisory, not worth their own threads

- **The lifecycle test can never pass again once the proposal executes on mainnet** — `:115-116` pin `proposalThreshold() == 5_000e18` and `quorumNumerator() == 3` against an unpinned latest-block fork. The rehearsal description correctly solves the *submission*-time `Governor: proposal already exists` collision, but not this. The documented runbook (re-run before submitting, again before executing) still works, since the threshold is unchanged until execution; what's lost is the run you'd most want afterwards — confirming the state actually landed. Reading the starting values rather than pinning them covers both.
- `:127` uses `assertGt` where the contract's rule is `quorum(snapshot) <= forVotes`. Strictly stronger than on-chain, and would fail on an exactly-at-quorum position the Governor accepts.
- `:145` tests "nothing in flight is affected" only against the *rehearsal* proposal, which is `EXECUTED` by that point, and whose snapshot sits ~32,728 blocks before the execution block — so the inclusive-lookup edge that actually decides the claim is never exercised, and no real in-flight proposal id is checked. `:147` returns 10 through OZ's optimistic latest-checkpoint branch, so it confirms nothing about the binary search.
- **Forge does not run in CI at all.** The workflow is eslint → solhint (globbed to `contracts/**`) → `hardhat test` → coverage, and Hardhat compiles only `contracts/`. So `.t.sol` and `.s.sol` here are neither compiled nor executed — "3/3 passing" has no automated re-verification, and this file could stop compiling with CI green. Pre-existing and repo-wide, not this PR's doing, but it removes the safety net under several of the notes above.
- `.s.sol:16-17` "held 5,100 OLAS, just over the 5,000 veOLAS minimum" conflates OLAS with veOLAS. It is true only because that proposer took a near-four-year lock (5,076 veOLAS); as written it implies an equivalence the token doesn't provide, and it's the motivating fact for the whole proposal. Same at `README.md:22`.
- The header block at `.s.sol:15-46` duplicates ~32 of its 42 lines from `proposal_15/README.md:19-55`, which is a strict superset (it also carries the id, the hash and the pointer to the analysis). Two copies of the flat-band wording to fix, rotting independently. This is a pre-existing house convention — 13 and 14 do the same — so it's a convention call rather than a regression; the difference is that 14's header is mostly mechanism while 15's is almost entirely live-figure argument, which doubles the stale surface.
- Live figures frozen into comments that nothing re-checks: "twelve addresses clear 5,000", "the sixth is below 16,000", "286,101 to 1,126,974", "Rank-1 veOLAS holder" at `:45`. The `assertGt` at `:119` will fail loudly when the DAO proposer's position inverts, but nothing checks the rank-5 position against 250,000, which is what the flat-band claim rests on.
- The file-level `/// @notice` block in the test sits above `interface IGovernor`, so NatSpec attaches to the interface rather than `Proposal15GovernorParamsTest`. `console2` is imported and unused. The two pure tests are gated behind a mandatory fork, where 13 and 14 fork inside only the bodies that need it — so the two checks an operator most wants before submitting can't run offline.
- One comment that earns its place: `:66-67`, on forking before deploying the builder. That's a real ordering constraint whose violation fails confusingly, and blame wouldn't recover it. Keep it.

Nice improvements over 11-14 worth noting: the rehearsal description, and not mocking voting power — 11 through 14 all `vm.mockCall` `getPastVotes`/`getPastTotalSupply` with synthetic actors, which structurally cannot detect a threshold or quorum no real holder can meet. That trade is the right one; it just needs the post-execution assertions to pay off.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233 — #232 merged while this was being written, so the follow-up carries all of these.

Now `vm.envOr("ETH_RPC", string("https://ethereum-rpc.publicnode.com"))` plus `assertEq(block.chainid, 1, "Must run on a mainnet fork")`, matching Proposal11/12/13/14. You are right that the documented `--fork-url` invocation failed `setUp` on this contract alone.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233. Added `assertEq(gov.hashProposal(t, v, c, keccak256(bytes(d))), PUBLISHED_ID, "id per the deployed Governor")` alongside the local recompute.

Your reasoning is the part worth keeping: in the local path `GOVERNOR` is only an array element, so a stale constant stays self-consistent while the real Governor assigns a different id. The analytics suite did assert this and it was dropped in the port.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233 — `test_committedArtifactsMatchTheBuilder`, same shape as Proposal13/14, and the builder now carries the byte-for-byte warning the other three have.

`foundry.toml` already granted `fs_permissions` for exactly this and I did not use it. The framing that made it land: `description.txt` is what a human pastes into `propose()`, Forge does not run in CI, so a stray newline had no second line of defence.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233, both halves.

Moved post-execution where it binds, and asserted with 2× headroom rather than at the boundary — your point that `proposalThreshold` is not checkpointed and the cancel gate reads it live means the assertion is *guaranteed* to invert as that lock decays toward 2027-01-07, and at the boundary it would first fail the day the cap is already breached.

Both directions are now behavioural: `testHostileProposalBecomesCancellable` (with a pre-change control) and `testLegitimateProposalStaysProtected`.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233 — added both assertions you wrote, post-execution.

Agreed on the severity framing: if 10% of supply exceeded every individual position, the only route back is a proposal that must clear the bar that broke it. That is the one failure mode here with no recovery, and it had no assertion.

scripts/proposals/proposal_15/Proposal15GovernorParams.s.sol
DavidMinarsch: Accepted in #233 — the claim was wrong and self-defeating exactly as you set out. The band is `(15,880 … 293,929]`, its top is 293,929, and "take the top of the band" would have selected that.

Builder and README now give the real reason: decay headroom plus the Bravo-cancel cap, with the ~15% gap stated as deliberate. As you note, the on-chain description never carried the band language, so nothing immutable is affected. The same wording exists in the analytics calibration note and I am fixing it there too.

scripts/proposals/proposal_15/README.md
DavidMinarsch: Fixed in #233. You are right that the turnout range undercut the conclusion: three of the five votes ever cast fall below a 10% bar, and they are the three lowest.

Rewritten to rest on position capacity, which stands on its own, and to state the 347,685 figure and that three of five miss it. Also updated for the post-relock reality — `P` is now 8,394,665, so a 10% quorum is 839,466, still cleared alone by each of the three largest positions at ~2.5M each. And I dropped the "15% would require two voters" line rather than defend it.

scripts/proposals/proposal_15/calldata.json
DavidMinarsch: Fixed in #233 — `annotate.js` and `proposal_15.html` added, plus the Files table and regenerate instructions the other folders carry. The generator step simply was not run.

Useful side effect: `annotate.js` recomputes the proposalId independently of both the builder and the test, and agrees — a third derivation of `3563000702…765109424`.

test/proposals/Proposal15GovernorParams.t.sol
DavidMinarsch: Fixed in #233. `console2` dropped; `hashProposal` and `cancel` are now called rather than declared and abandoned — your reading of them as a map of intended-then-dropped coverage was exactly right.

PR #233, opened 2026-09-07
proposal 15: real assertions, the missing artifacts, and two prose fixes

DavidMinarsch:
Follow-up to #232, which merged while this was being written. Same folder, no change to the payload, the description or the proposalId — the proposal is on-chain and unaffected. This is the review on #232 applied.

## The test suite was weaker than it read (3 tests → 6)

- **Fork config** now matches every other proposal test: `vm.envOr("ETH_RPC", "https://ethereum-rpc.publicnode.com")` plus `assertEq(block.chainid, 1)`. The old `vm.envString("ETH_RPC_URL")` used a variable no other suite uses and had no default, so `forge test --fork-url …` — the invocation documented in `CLAUDE.md` — failed `setUp`.
- **The published id is now bound to the deployed Governor**, via `gov.hashProposal(...)`, not only to a local recompute. `GOVERNOR` appears in the local path only as an array element, so a stale constant would still produce a self-consistent pass while the real Governor assigned a different id.
- **`test_committedArtifactsMatchTheBuilder`**, matching proposals 13 and 14. `foundry.toml` already granted `fs_permissions` for exactly this and this folder wasn't using it. `description.txt` is what a human pastes into `propose()`; a stray trailing newline there changes the id and nothing would have caught it.
- **The cap is asserted post-execution, where it binds, with 2× headroom.** `proposalThreshold` is *not* checkpointed and Bravo's cancel gate reads it live, so the assertion is guaranteed to invert as the DAO proposer's power decays toward its 2027-01-07 lock end. At the boundary it would first fail on the day the cap is already breached; with headroom it fails while there is time to react.
- **Both directions of the cancel guard are behavioural now**, not a single view call: the hostile 2026-09-05 proposal becomes cancellable by anyone after execution (with a pre-change control so it can't pass vacuously), and proposal 13 does not.
- **Governance still works after the change** — the largest holder is asserted to clear both the new threshold and the new quorum. If 10% of supply exceeded every individual position, the only route back would be a proposal that must clear the bar that broke it.
- Dropped the unused `console2`; `hashProposal` and `cancel` are now used rather than declared and abandoned.

## Two prose defects

**"250,000 is the top of a flat band" was wrong.** The band is `(15,880 … 293,929]` and its top is 293,929 — by the stated rule the choice would have been that, ~15% higher, admitting the identical five addresses. The actual reason for 250,000 is decay headroom plus the cancel cap, so the builder and README now say that. The on-chain description never carried the band language, so nothing immutable is affected.

**The quorum paragraph cited turnout that undercut it.** Three of the five votes ever cast fall below a 10% bar. The case rests on *position capacity*, not past turnout, and now says so — with the post-relock figures (`P` 8,394,665, 10% quorum 839,466, each of the three largest positions ~2.5M).

## Missing artifacts

`annotate.js` and `proposal_15.html`, required of every proposal folder by `scripts/proposals/README.md` and shipped by 11–14. The generator recomputes the proposalId independently of both the builder and the test, and agrees: `3563000702947626407249966116394330391891319244018918786718598572660765109424`. README gains the Files table and regenerate instructions the other folders carry.

## Checks

`ETH_RPC=<rpc> forge test --match-contract Proposal15` — 6/6 passing.


OjusWiZard (approved):
## Approve

This does what #232 asked for, and the important property holds: **the submitted payload is untouched.** `description.txt` and `calldata.json` are byte-identical to before, the builder's `abi.encodePacked` block is unchanged (both hunks in that file are comment-only), and the rebuilt description is still 591 bytes matching the committed file exactly. The proposal on-chain is unaffected by anything here.

The band correction, the turnout paragraph now stating plainly that three of the five votes ever cast fall below a 10% bar, `test_committedArtifactsMatchTheBuilder`, the `ETH_RPC` alignment, and both cancel-guard directions are all real improvements. Approving on that basis.

One thing that is not about this PR's code, but which this PR's contents are now measured against, and which is time-sensitive.

---

### The 250,000 threshold no longer achieves the cancel-guard purpose it is justified by

Verified against mainnet at block **25,927,577** via `ethereum-rpc.publicnode.com`:

| | |
|---|---:|
| Hostile proposer `0xd4f8d1AE…5451`, `wveOLAS.getVotes` | **116,255 veOLAS** |
| New threshold | 250,000 |
| Gap | **133,744** |
| OLAS to close it at their existing ~2030-08-29 lock (η ≈ 0.995) | ~134,416 |
| Cost at $0.0266/OLAS | **~$3,575** |

Every artifact in this folder still describes that proposer as holding **5,074 veOLAS** — `Proposal15GovernorParams.s.sol:16-17` and `:50-51`, `README.md:22`, and the test's constant NatSpec. That figure is **22.9× stale**. The change is not decay; it is deliberate accumulation since 2026-09-05.

The consequence is specific. `README.md:39-44` and the builder header at `:27-31` justify the parameter partly on Bravo's permissionless `cancel` opening against a minimum-stake proposer. At 116,255 they are 2.15× under the bar, so the guard *would* bind if executed today — but the margin is about $3,600 of OLAS and closing it is entirely their choice, at any moment before execution. The half of this proposal that is aimed at the proposer who prompted it is purchasable for roughly what the original attempt cost.

That is a decision rather than a defect, and it belongs to the DAO: execute anyway and accept the guard is defeatable, or retire the cancel-guard justification and rest the case on the quorum change, which is unaffected and is the stronger half regardless. It has to be settled before execution — the parameter is fixed in the submitted bytes, so revising it means voting 15 down and resubmitting. Proposal 15 is `Pending` as of this writing (`proposalThreshold()` still 5,000e18, numerator still 3), so there is time.

For completeness, three related figures I checked at the same block, since they bear on the same paragraphs:

- **`veOLAS.lockedEnd(0x3447…0039)` is 2027-09-02, not 2027-01-07.** The 2026-09-07 relocking extended it. The comment at `Proposal15GovernorParams.t.sol:194-195` and the cliff arithmetic in the companion analytics note both derive decay dates from January; all of them are wrong, in the safe direction.
- **The cap is comfortable and was never at risk** — the DAO proposer holds 2,460,191, giving 9.84× headroom over 250,000, well past the 2× the test asserts.
- **`P` is 8,387,545**, within 0.1% of the 8,394,665 quoted at `README.md:42`, so the post-relock figure is sound. It also confirms the pre-relock set at `README.md:25-27` (982,733 / 856,538 / 838,898 / 409,732 / 293,929, sixth at 15,880, summing to 3.4M) is superseded — those are still stated in the present tense two paragraphs above the number that supersedes them, and the band argument rests on them. A 10% numerator now puts quorum at 838,755. I could not enumerate holders to recheck "twelve addresses clear 5,000, five clear 250,000" without an Etherscan key, so whether that sentence in the immutable description is still accurate is open.

The remaining items from my #232 review that this PR did not close — the `annotate.js` copy still carrying no proposal-15 selectors so `proposal_15.html` renders both calls as `unknown` / `unrecognized selector`, the `:202` quorum assertion reading the pre-checkpoint 3% numerator, the two tests that go red once 15 executes, and the missing propose-gate test — are all worth doing but none of them gates execution. Happy to open a follow-up issue for them rather than hold this.

One correction of my own: the rationale I gave on #232 for adding `gov.hashProposal`, now quoted in the code at `:102-105`, is wrong. `hashProposal` is `pure` and takes no Governor identity, so a stale `GOVERNOR` constant shifts the local recompute and the on-chain call identically — what catches that is the hardcoded `PUBLISHED_ID` on the line above. The assertion is still worth keeping (it pins the id *scheme* to the deployed bytecode), but the comment should not be cited as precedent.

PR #234, opened 2026-09-11
proposal 16: adopt the Olas instance on Robinhood Chain as canonical

DavidMinarsch:
Asks the DAO to recognise the existing Robinhood Chain (4663) deployment as **the canonical Olas instance on that chain**, and connects it to the protocol on L1 accordingly.

**proposalId** `0xe8f681809e73bf7044bc0c830f30cbd50a02ca0b56677fbef742b4a66dd4ede4`
**descriptionHash** `0x0c6c2c6f9a6f27dcf1f7a14fb197e788de136f50f36d177e031713e428cc3c9d`

## What the vote is about

A complete Olas protocol instance already exists on 4663 — registries, service management, staking factory and verifier, mech marketplace, buy-back stack. All 28 contracts are verified on Sourcify, and on 2026-09-11 every owner-bearing contract was transferred to the chain's governance control point, so **this DAO already owns it**.

What it is *not* yet is part of the protocol. The L1 contracts do not recognise the chain, so no staking incentives can reach it and the community multisig cannot act on it. This proposal is the DAO adopting that deployment as canonical; the three calls are what that adoption consists of on L1.

| # | Target | Call |
|---|---|---|
| 0 | `Dispenser` `0x5650300f…` | `setDepositProcessorChainIds([0xb9DfcC61…], [4663])` |
| 1 | `GuardCM` `0xC0b146D6…` | `setBridgeMediatorL1BridgeParams([inbox], [verifier], [4663], [mediator])` |
| 2 | `GuardCM` `0xC0b146D6…` | `setTargetSelectorChainIds(…)` — `pause()` ×2, `drain()`, `drain(address)` |

All three are **direct L1 Timelock calls** — nothing is bridged, so every effect is observable on a mainnet fork and there is no destination-chain leg to simulate.

## Entry 2 is the fleet's selector set, verified not assumed

| target | selector | on the other seven chains |
|---|---|---|
| `ServiceManagerProxy` | `pause()` | `true` |
| `ArbitrumTargetDispenserL2` | `pause()` | `true` |
| `ServiceRegistryL2` | `drain()` `0x9890220b` | `true` |
| `ServiceRegistryTokenUtility` | `drain(address)` `0xece53132` | `true` |
| either pausable target | `unpause()` | **`false` — not granted** |

`unpause()` was in an earlier revision and removed after @OjusWiZard's review. It is allowlisted on no chain, and granting it would let the CM lift a pause on 4663 in one transaction with no vote — the CM holds the Timelock's `PROPOSER_ROLE` and `EXECUTOR_ROLE`, `getMinDelay()` is `0`, and GuardCM is the CM Safe's guard. The `drain` selectors were added for the same parity reason in the opposite direction.

**Entries 1 and 2 ship together deliberately.** Mode's allowlist was backfilled in July 2026 without `setBridgeMediatorL1BridgeParams`, leaving `verifierL2 == 0` and every CM transaction to Mode failing closed for two months until proposal 13.

## Address collision — expected

The mediator `0x4d30F68F…` is `alias(L1 Timelock)` and is byte-identical to Arbitrum One's; the verifier `0x0F336366…` serves every Orbit chain. Only the inbox and chain Id distinguish the routes — so an accidental overwrite of Arbitrum's entry would leave two of three fields looking right. The fork test asserts it is untouched, chain Id included.

## Validation

```
forge test --match-contract Proposal16RobinhoodTest -vv
```

- `test_preconditions` — everything this proposal sets is unset first, with the populated Arbitrum route as a positive control, plus the fleet-parity assertions that `unpause()` is granted nowhere.
- `test_L1_fullGovernanceLifecycle` — propose → vote → queue → execute through the **live** GovernorOLAS; asserts the governor assigns the published `proposalId`, then all three effects and the Arbitrum non-regression.
- `test_CM_canPause4663_onlyAfterProposal` — what the guard **admits**, not just what it stores: a real CM `schedule` carrying an Orbit retryable, rejected before with a pinned `NotAuthorized(inbox, createRetryableTicket, 1)`, accepted after for both pausable targets, `unpause()` still refused.

All three pass; `execute()` ~266k gas. `annotate.js` recomputes the id in ethers independently of Solidity and the two agree; `description.txt` matches the builder constant byte-for-byte.

## ⚠️ Note for re-reviewers

The `proposalId` has changed **twice** since the first approvals, both times because the description changed:

| | proposalId |
|---|---|
| original (approved by @dagacha, reviewed by @OjusWiZard) | `0x8a62ccd8…` |
| after dropping `unpause`, adding `drain` | `0x5404dbff…` |
| **current**, after the adoption reframing | **`0xe8f68180…`** |

Targets, values and calldatas are unchanged from the second revision — `calldata.json` is byte-identical — but the id is hashed over the description, so the earlier approvals were against different bytes.

## Numbering

This is **16**, not 15 — `proposal_15` is the Governor threshold/quorum change already on `main`. Separately: proposals **14 and 15 are missing from the index table**, which now reads 13 → 16. I added only my own row.


scripts/proposals/proposal_16/Proposal16Robinhood.s.sol
```diff
@@ -0,0 +1,195 @@
...
+    function _guardCmRobinhoodSelectors() internal pure returns (bytes memory) {
+        address[] memory targets_ = new address[](4);
+        targets_[0] = ROBINHOOD_SERVICE_MANAGER_PROXY;
+        targets_[1] = ROBINHOOD_SERVICE_MANAGER_PROXY;
+        targets_[2] = ROBINHOOD_TARGET_DISPENSER_L2;
+        targets_[3] = ROBINHOOD_TARGET_DISPENSER_L2;
+
+        bytes4[] memory selectors = new bytes4[](4);
+        selectors[0] = SELECTOR_PAUSE;
+        selectors[1] = SELECTOR_UNPAUSE;
```
OjusWiZard: **This gives the community multisig (CM) `unpause()` on 4663, which it has on no other chain.**

Live `GuardCM.getTargetSelectorChainId` today, on gnosis, polygon, arbitrum, optimism, base, celo and mode: `pause()` is `true` and `unpause()` is `false`, for both `ServiceManagerProxy` and the L2 target dispenser. `pause()` reads `true` on the same address and key, so the `false` is real. Mainnet matches: Treasury `pause()` is `true`, `unpause()` is `false`.

The grant takes effect immediately. The CM holds the Timelock's `PROPOSER_ROLE` and `EXECUTOR_ROLE`, `getMinDelay()` is `0`, and the CM Safe's guard is this GuardCM, so the allowlist is the only constraint. After execution, the CM could lift a pause on either 4663 contract in a single transaction, with no vote.

The description, which is hashed into the `proposalId`, calls entry 2 "the same fast-path containment it holds on the other chains". Either drop the two `unpause` triples (`targets_[1]` and `[3]`), or keep them and state in the description that 4663 gives the CM a power it lacks elsewhere. Either way, `calldata.json`, the `proposalId`, the HTML and the test all need regenerating.

scripts/proposals/proposal_16/Proposal16Robinhood.s.sol
```diff
@@ -0,0 +1,195 @@
...
+        address[] memory l2s = new address[](1);
+        l2s[0] = ROBINHOOD_MEDIATOR_L2;
+        return abi.encodeWithSignature(
+            "setBridgeMediatorL1BridgeParams(address[],address[],uint256[],address[])", l1s, vs, cids, l2s
+        );
+    }
+
+    /// @dev GuardCM.setTargetSelectorChainIds — four triples, all enabled, all on 4663.
+    ///      Order is (ServiceManagerProxy, dispenser) x (pause, unpause).
+    function _guardCmRobinhoodSelectors() internal pure returns (bytes memory) {
```
OjusWiZard, line 159: Minor, for parity: every other L2 also allowlists `ServiceRegistryL2.drain()` (`0x9890220b`) and `ServiceRegistryTokenUtility.drain(address)` (`0xece53132`) for the CM. Proposal 11 set them, and both read `true` on all seven chains today. Draining isn't containment, so leaving them out doesn't contradict the description. But without them, 4663 would be the only chain where draining slashed funds needs a full governance cycle. Worth adding (`0xE3607b00…`, `0x3d77596b…`), or noting that it's deliberate.

test/proposals/Proposal16Robinhood.t.sol
```diff
@@ -0,0 +1,171 @@
...
+            "entry 0: 4663 processor not registered"
+        );
+
+        // [1] GuardCM holds the 4663 route. verifierL2 != 0 is the exact check Mode failed.
+        (address v, address m, uint256 c) = IGuardCM(GUARD_CM).mapBridgeMediatorL1BridgeParams(ROBINHOOD_INBOX);
+        assertEq(v, VERIFIER_ARBITRUM, "entry 1: wrong verifier");
+        assertEq(m, ROBINHOOD_MEDIATOR_L2, "entry 1: wrong mediator");
+        assertEq(c, CID_ROBINHOOD, "entry 1: wrong chainId");
+
+        // [2] all four triples are allowlisted for 4663.
```
OjusWiZard: Low: these assertions check what is stored, not what the guard actually lets through. Proposal 13's test also runs `GuardCM.checkTransaction` on a real CM `schedule`, which is the Mode lesson this PR cites.

Nothing is broken: I planted this proposal's state into GuardCM with an override, and `schedule` → `createRetryableTicket` → `pause()`/`unpause()` passes for both targets, while live state rejects it with `NotAuthorized`. A capability test would just keep that tested.

OjusWiZard (commented):
I checked the calldata, both hashes and the effects independently: decoded and re-encoded all three entries byte-for-byte; the live governor's `hashProposal` returns the published id; all three entries succeed as Timelock calls; and a simulated GuardCM `checkTransaction` accepts the 4663 route. One thing to resolve before submission, inline on the builder, plus two minor notes.

dagacha (approved):
L1 wiring and calldata are correct. Independently checked on mainnet and 4663: official Delayed Inbox (`bridge()`/`rollup()` match published contracts), processor `l2TargetChainId()==4663` with dispenser/inbox/gateway router lined up, SMP and L2 dispenser `owner()==alias(Timelock)`, GuardCM Arbitrum slot untouched. `description.txt` hashes to `0xf9a031f3…` (no trailing newline); HTML `proposalId` `0x8a62ccd8…`. `forge test --match-contract Proposal16RobinhoodTest -vv` — both tests pass, `Governor.execute()` ~266k gas.

Shared Orbit verifier and alias are load-bearing, not copy-paste.

Review: Omp · grok-4.6 · effort Medium

dagacha (commented):
Reviewed commit `752bfbe56a5c4daaa74d845c6dde8cdbb9dfb866` with Codex (LLM: GPT-6; effort: Medium).

No actionable findings.

Validation performed:
- Both `Proposal16RobinhoodTest` mainnet fork tests passed.
- The committed calldata and description match the Solidity builder byte-for-byte, and the independently calculated proposal ID agrees.
- Read-only L1/L2 RPC checks confirmed the processor wiring and governance ownership of both allowlisted L2 contracts.
- The fork test confirmed that the existing Arbitrum route remains intact.

Validation limits: the lifecycle test mocks voting power; end-to-end cross-chain delivery was not simulated.

There are no inline comments because no actionable line-specific issues were identified.

dagacha (dismissed):

OjusWiZard (approved):
Approving.

dagacha (commented):
Follow-up review of `700d79a12f7c4b35efc4c62517ffbcc2f5880a41`, covering changes since `752bfbe`, performed with Codex (LLM: GPT-6; effort: Medium).

No new actionable findings.

The removal of `unpause()` corrects the permission expansion identified in the intervening review. My initial review missed that mismatch with the existing chain permissions. The updated selector set contains two `pause()` grants and the two registry drain grants.

Validation:
- All three `Proposal16RobinhoodTest` mainnet fork tests pass, including the new before/after GuardCM admission test and rejection of `unpause()`.
- Both new L2 drain targets report the Timelock alias as owner and drainer. Read-only `eth_call` simulations from that alias succeed for `drain()` and `drain(OLAS)` at current state (both return zero).
- All three calldata entries and the description match the Solidity builder byte-for-byte. The independently recomputed proposal ID is `0xe8f681809e73bf7044bc0c830f30cbd50a02ca0b56677fbef742b4a66dd4ede4`.
- The HTML regenerates byte-identically with its committed title, and ESLint passes for `annotate.js`.

Limits: voting power remains mocked; the new CM test checks GuardCM admission, not actual bridge delivery. The read-only drain simulations do not exercise a nonzero fund transfer.

dagacha (approved):
Follow-up on 700d79a. Entry 2 now matches the live fleet: pause×2 + drain×2, no unpause. Independently checked on 4663 that SRL2 0xE3607b00… and SRTU 0x3d77596b… exist and owner()/drainer() are alias(Timelock). Live GuardCM: unpause false on Arbitrum/Mode; drain true on Arbitrum/Mode/Polygon. description.txt = builder (0x0c6c2c6f…); HTML proposalId 0xe8f68180…. forge test --match-contract Proposal16RobinhoodTest -vv — 3 pass, including the CM schedule capability test (rejected before with pinned NotAuthorized, pause accepted after, unpause still refused).

This id is not the one previously approved (0x8a62ccd8…).

Review: Omp · grok-4.6 · effort Medium

dagacha (approved):

scripts/proposals/proposal_16/Proposal16Robinhood.s.sol
DavidMinarsch: Actioned — both `unpause` triples dropped. Verified your claim before changing anything: `unpause()` reads `false` on gnosis, polygon, arbitrum, optimism, base, celo and mode for both `ServiceManagerProxy` and the L2 dispenser, with `pause()` `true` on the same key, so the `false` is real. (My first celo read said `false` for `pause()` too — my SMP address was wrong; with `0x84B4DA67…` it is `true`/`false` like the rest.)

Entry 2 is now the fleet set: `pause()` on the two pausable targets, `drain()` / `drain(address)` on the two registries. The description no longer claims parity it did not have, and it now states the exclusion explicitly.

The test asserts the negative too — both `unpause` triples must still read `false` after execution — so a future edit reintroducing them fails the test. Worth knowing, and flagged separately: governance CI runs eslint/solhint/hardhat only, no Forge, so that test is not actually run by CI and has to be run by hand.

New id `0xe8f68180…` (it moved twice: once here, once for a description reframing).

scripts/proposals/proposal_16/Proposal16Robinhood.s.sol
DavidMinarsch: Added. Confirmed first: `ServiceRegistryL2.drain()` (`0x9890220b`) and `ServiceRegistryTokenUtility.drain(address)` (`0xece53132`) read `true` on all six chains I could check, set by proposal 11.

You were right that this cuts the other way from the `unpause` finding — leaving them out would have made 4663 the only chain where draining slashed funds needs a full governance cycle, which is a parity gap in the opposite direction. Entry 2 is still four triples, just a different four.

Also recorded the general rule in the rollout plan so the next chain derives this set by querying live state rather than by judgement, since both findings came from the same root cause.

test/proposals/Proposal16Robinhood.t.sol
DavidMinarsch: Added, having first checked the precedent you pointed at — proposal 13 does run `GuardCM.checkTransaction` on a real CM `schedule`, so this follows it rather than inventing a shape.

`test_CM_canPause4663_onlyAfterProposal` builds a CM `schedule` carrying an Orbit retryable to 4663 and runs it through the guard: rejected before with a pinned `NotAuthorized(inbox, createRetryableTicket, 1)` rather than a bare `expectRevert`, accepted after for both pausable targets, and `unpause()` still refused. That last leg is what makes it cover the corrected allowlist as well as the route.

Your framing was the useful part: the stored-state assertions could not have caught the Mode failure mode this proposal cites, where storage said yes and the guard said no.

PR #236, opened 2026-09-22
governance review: activation, roles, proposal inventory and contract matrix

mariapiamo:
Records the first completed part of the governance review covering the source range `v1.2.5-post-external-audit` → `v1.2.5` → `v1.3.0-pre-external-audit`, and on-chain governance activity from May 1, 2026. Ethereum state is pinned at block 26,032,428. Every claim links to saved evidence, and the collectors and tests are reproducible from the repository root.

**Scope.** Governance contracts and the authority they hold over other systems. External business logic is not reviewed. This is not a completed audit.

## What is established

- **Activation is verified, not merely documented.** The new Governor holds the Timelock admin, proposer, executor and canceller roles; the previous Governor holds none. The CM Safe holds proposer and executor only, and its guard was switched in the June 15 activation transaction. Reconstructed from 31 Timelock role events spanning deployment to the pin, reconciled against `hasRole` and `getRoleAdmin`.
- **Timing parameters.** Voting delay 13,091 blocks, voting period 19,636 blocks, Governor execution delay 157,092 seconds. The Timelock's own minimum delay is zero, so the execution delay is carried per operation by the Governor; a queued proposal cannot be executed early by any role holder.
- **Proposal inventory.** Twelve proposals in the window: nine executed, one cancelled, two defeated. Proposals 12–16 match their repository artifacts exactly in description and in ordered target/value/calldata. Defeat is inferred from ended voting with zero For and positive Against votes.
- **Traceability gap on proposal 11.** The executed proposal carries 22 calls; the committed artifact has 16. Calls 0–13 and 15 match, call 14 differs, and calls 16–21 — ServiceManager upgrades across seven networks — are absent from the artifact. Four further executed or cancelled proposals have no artifact in this repository. This is a reproducibility gap, not evidence that the executed actions were wrong.
- **Contract matrix.** Source change per tag interval, deployed instance, live status, verified source version and build settings for every repository contract, with each status tied to the live pointer that establishes it. Net production logic changes occur only in GuardCM and VoteWeighting. The deployed GuardCM includes the zero-address check added in the first interval and differs from the repository source only in the second interval's comment block, verified by a comment-aware comparison. The deployed Timelock inherits OpenZeppelin `TimelockController` v4.6.0 against v4.8.2 in the current dependency, so a Timelock built from today's source is not the live contract.
- **VoteWeighting's revised version is not live.** The active instance is `0x95418b46…`, whose verified source matches the pre-May baseline exactly. No revised deployment was found in either repository's configuration or by explorer search; absence is not proven.

## Follow-up

Not completed in this repository, and continuing as follow-up work:

- **Voting power** — holder distribution, who meets the proposal threshold, and quorum coverage.
- **VoteWeighting** — source-fix review and the preconditions for a future deployment.
- **GuardCM** — configuration and cross-chain permissions, including the routes added during the window.
- **Veto system** — deployment status and design review.

The checklist records the open items and their stopping rules.

## Reproduction

See `audits/2026-09-22-governance-review/README.md`. Offline consistency check: `python3 audits/2026-09-22-governance-review/check_summary_evidence.py` (13/13).


audits/2026-09-22-governance-review/SUMMARY.md
```diff
@@ -0,0 +1,108 @@
+# Governance review — progress checklist
+
+**As of September 23, 2026.** Checked boxes mean the stated work is supported by saved evidence; unchecked boxes mean it remains open. Main governance state is pinned to Ethereum block **26,032,428**; the proposal inventory ends at **26,032,401**. This checklist is not a completed audit. VoteWeighting deployment/wiring was additionally checked at block **26,034,116** (September 22, 16:07:11 UTC). Holder eligibility and voting-power distribution were checked at finalized block **26,038,740** (September 23, 07:37:59 UTC); other state claims retain their original pins. **References on checked items support the recorded result; references on unchecked items identify review inputs or documented gaps, not proof that the work is complete.**
```
dagacha: Medium. This sentence says holder eligibility and voting-power distribution were checked at block 26,038,740. No pin, hash, or result for that block is in the PR. Line 34 leaves the same work unchecked, and check_summary_evidence.py skips it on purpose. Delete the sentence. Do not cite a block that has no saved evidence.

audits/2026-09-22-governance-review/build_contract_matrix.cjs
```diff
@@ -0,0 +1,265 @@
...
+   }
+  }
+  rows.push(row);
+ }
+ // Active status, each tied to the pointer that establishes it.
+ const status=(row)=>{
+  const a=(row.address||'').toLowerCase();
+  if(!row.address)return {status:'not deployed',evidence:'No address in mainnet configuration'};
+  if(row.contract.startsWith('Veto'))return {status:'not deployed',evidence:'No veto address in mainnet configuration'};
+  if(row.contract==='GovernorOLAS (current)')return {status:'active',evidence:`Timelock roles held; guard.governor()=${pointers.guardGovernor}`};
```
dagacha: Medium. status() does not use the pointer reads it already made. Old Governors are labeled "Holds no Timelock role" by contract name. Current Governor, GuardCM, and VoteWeighting are labeled active without requiring the pointer to equal the row address. call() returns an error object on failure, and that object is still interpolated into an active/superseded label. Today's JSON matches the pointers, so the committed matrix is not false. A regeneration would not fail if it were. Derive each status from the pointer, and throw if a read errors or the address does not match.

audits/2026-09-22-governance-review/build_contract_matrix.cjs
```diff
@@ -0,0 +1,265 @@
...
+ if(commentOnly.length)md.push('- Verified source matches no tracked version only because of comments; the executable source is identical to the working tree once comments and whitespace are removed: '
+  +commentOnly.map(label).join(', ')+'.');
+ for(const r of partial){
+  const d=r.diffVsCheckout;
+  const how=d&&d.codeIdenticalIgnoringComments
+   ? 'the only differences from the working tree are comments: the executable source is identical once comments and whitespace are removed'
+   : 'the working-tree source differs in code, not only in comments';
+  md.push(`- ${r.contract} (\`${r.address}\`) matches ${r.matchesSourceVersions.join(', ')}; ${how}.`);
+ }
+ md.push(`- Every verified runtime listed above matches the bytecode returned by RPC at the pin (${rows.filter(r=>r.explorerRuntimeMatchesRpc===true).length} of ${rows.filter(r=>r.address).length} deployed entries).`);
```
dagacha: Medium. This always writes "Every verified runtime listed above matches…". Only the count changes. If a later run has a mismatch, the note still states a universal pass. Throw when the true-count is not the deployed-count, or emit the sentence only in that case.

dagacha (commented):
Reviewed with Omp. LLM: Claude. Effort: Medium.

The checked role, proposal, and VoteWeighting conclusions match the committed evidence. The default check_summary_evidence.py run is 13/13. That is recorded-JSON and link consistency, not a chain proof.

Six medium issues.

The PR description overstates the delay. "A queued proposal cannot be executed early by any role holder" is not what was verified, and getMinDelay() == 0 does not imply it. SUMMARY.md line 32 already records the live property: CM holds proposer and executor and can schedule and execute a guard-permitted action without the Governor cycle. A Governor-queued operation carries its own eta (governorDelay = 157,092 seconds). Rewrite that sentence to those two facts. scripts/proposals/proposal_16/README.md states the same constraint for the allowlist.

audits/2026-09-22-governance-review/build_contract_matrix.cjs
```diff
@@ -0,0 +1,265 @@
...
+ ['OptimismMessenger','contracts/bridges/OptimismMessenger.sol',globals.baseMessengerL2Address,'Base'],
+ ['OptimismMessenger','contracts/bridges/OptimismMessenger.sol',globals.celoMessengerL2Address,'Celo'],
+ ['OptimismMessenger','contracts/bridges/OptimismMessenger.sol',globals.modeMessengerL2Address,'Mode'],
+ ['FxERC20ChildTunnel','contracts/bridges/FxERC20ChildTunnel.sol',globals.fxERC20ChildTunnelAddress,'Polygon'],
+ ['WormholeMessenger','contracts/bridges/WormholeMessenger.sol',null,'Wormhole route (no address in mainnet configuration)'],
+ ['WormholeRelayerTimelock','contracts/bridges/WormholeRelayerTimelock.sol',null,'Wormhole route (no address in mainnet configuration)']];
+
+async function main(){
+ if(await rpc('eth_chainId',[],false)!=='0x1')throw Error('Wrong chain');
+ const header=await rpc('eth_getBlockByNumber',[hex(pin.number),false]);
```
dagacha: Medium. The pin-hash check is cached. The chain-id read correctly bypasses the cache; this block-header read does not. A re-run compares pin.json to the header saved by the last successful run, so it cannot detect a reorg or a stale cache. Bypass the cache here. Same fix at point-1/collect_role_history.cjs line 37 and point-3/verify_deployment.cjs line 69.

audits/2026-09-22-governance-review/check_summary_evidence.py
```diff
@@ -0,0 +1,101 @@
...
+check('All report section references resolve', not bad_anchors)
+items = [line for line in (ROOT / 'SUMMARY.md').read_text().splitlines() if re.match(r'- \[[ x]\] ', line)]
+check('Every checklist item has a reference', all(re.search(r'\]\([^)]+\)', line) for line in items))
+raw_results = []
+if args.check_raw:
+    for name, expected_hash in evidence['inputSha256'].items():
+        path = ROOT / name
+        matched = path.exists() and hashlib.sha256(path.read_bytes()).hexdigest() == expected_hash
+        raw_results.append({'file': name, 'matches': matched})
+    for name, expected_hash in deployment['localInputSha256'].items():
```
dagacha: Medium. --check-raw references deployment without defining it, so this path raises NameError before it hashes anything. The default 13/13 run does not hit it. Load the deployment evidence (or skip that loop) before indexing localInputSha256.

